PR #17, opened 2020-01-24
Use typescript on the frontend

kodyang:
### AB Task
https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/33288

### Overview
- Added typescript to packages.
- Modified existing file endings from `js` to `tsx`
- Added types to existing files the best I could.


private-templates-service/client/src/Config.tsx
```diff
@@ -0,0 +1,7 @@
+import { } from "react"; // Typescripte requires every file to have an import
+
+export default {
+  appId: "4803f66a-136d-4155-a51e-6d98400d5506",
```
daniellemott: I think app registration should be fixed or removed from this PR. What do you think?

gracez72: ^ I agree - we should also start setting client ids as secret

private-templates-service/client/src/ErrorMessage.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Alert } from "reactstrap";
+
+interface Props {
```
Jackbk: Might want to think about what naming conventions you are going to start using.  Maybe 'ErrorMessageProps', since this will be able to be accessed outside this file

private-templates-service/client/src/ErrorMessage.tsx
```diff
@@ -0,0 +1,26 @@
...
+      debug = (
+        <pre className="alert-pre border bg-light p-2">
+          <code>{this.props.debug}</code>
+        </pre>
+      );
+    }
+    return (
+      <Alert color="danger">
+        <p className="mb-3">{this.props.message}</p>
+        {debug}
```
Jackbk: You can simplify the if statement to 
`{ debug && <pre className="alert-pre border bg-light p-2">
          <code>{this.props.debug}</code>
        </pre>}`

private-templates-service/client/src/GraphService.tsx
```diff
@@ -1,27 +1,27 @@
-var graph = require('@microsoft/microsoft-graph-client');
+var graph = require("@microsoft/microsoft-graph-client");
```
Jackbk: Types in this class need to be defined too

private-templates-service/client/src/GraphService.tsx
```diff
@@ -1,27 +1,27 @@
...
       done(null, accessToken.accessToken);
     }
   });
 
   return client;
 }
 
-export async function getUserDetails(accessToken) {
+export async function getUserDetails(accessToken: any) {
   console.log(accessToken);
```
Jackbk: Avoid logging sensitive data like access tokens (also if you can add a logging framework your life will become nicer later on)

private-templates-service/client/src/GraphService.tsx
```diff
@@ -1,27 +1,27 @@
...
   const client = getAuthenticatedClient(accessToken);
```
Jackbk: Should GraphService be a class, so you can store this in state and not create a new one each time its called?

private-templates-service/client/src/NavBar.tsx
```diff
@@ -11,41 +11,54 @@ import {
...
   DropdownToggle,
   DropdownMenu,
-  DropdownItem } from 'reactstrap';
-import '@fortawesome/fontawesome-free/css/all.css';
+  DropdownItem
+} from "reactstrap";
+import "@fortawesome/fontawesome-free/css/all.css";
 
-function UserAvatar(props) {
+function UserAvatar(props: any) {
```
Jackbk: While implicit typing is nice, I find a good habit to get into is having functions return an explicit type for safety (for the next person who comes along and tries to modify it to return a number for example, they will know that other places expect it to be a component)

`function UserAvatar(props: any): Component`

private-templates-service/client/src/NavBar.tsx
```diff
@@ -11,41 +11,54 @@ import {
...
-    return <img
-            src={props.user.avatar} alt="user"
-            className="rounded-circle align-self-center mr-2"
-            style={{width: '32px'}}></img>;
+    return (
+      <img
+        src={props.user.avatar}
+        alt="user"
+        className="rounded-circle align-self-center mr-2"
+        style={{ width: "32px" }}
```
Jackbk: Bring styles out into its own file

private-templates-service/client/src/NavBar.tsx
```diff
@@ -57,8 +70,8 @@ function AuthNavItem(props) {
   );
 }
 
-export default class NavBar extends React.Component {
-  constructor(props) {
+export default class NavBar extends React.Component<any, any> {
+  constructor(props: any) {
```
Jackbk: NavBarProps

private-templates-service/client/src/NavBar.tsx
```diff
@@ -84,13 +97,16 @@ export default class NavBar extends React.Component {
               <Nav className="ml-auto" navbar>
                 <AuthNavItem
                   isAuthenticated={this.props.isAuthenticated}
-                  authButtonMethod={this.props.authButtonMethod}
-                  user={this.props.user} />
+                  authButtonMethod={
```
Jackbk: Should this component have a render method?

private-templates-service/client/src/Welcome.tsx
```diff
@@ -1,15 +1,22 @@
-import React from 'react';
-import {
-  Button,
-  Jumbotron } from 'reactstrap';
+import React from "react";
+import { Button, Jumbotron } from "reactstrap";
 
-function WelcomeContent(props) {
+interface Props {
```
Jackbk: WelcomeProps

private-templates-service/client/src/Welcome.tsx
```diff
@@ -1,15 +1,22 @@
...
+  isAuthenticated: boolean;
+  user: any;
```
Jackbk: Types for user and authButtonMethod

private-templates-service/client/src/Welcome.tsx
```diff
@@ -1,15 +1,22 @@
...
+  authButtonMethod: any;
+}
+
+function WelcomeContent(props: any) {
```
Jackbk: props: WelcomeContentProps

private-templates-service/client/src/Welcome.tsx
Jackbk: I'm seeing this structure a lot, it might be worth creating its own type to contain this data if you dont use a global store

private-templates-service/client/src/Welcome.tsx
```diff
@@ -1,15 +1,22 @@
...
   // If authenticated, greet the user
   if (props.isAuthenticated) {
     if (props.user.organization) {
       return (
         <div>
-          <h4>Welcome {props.user.displayName} from {props.user.organization}!</h4>
+          <h4>
+            Welcome {props.user.displayName} from{" "}
+            {props.user.organization}!
+					</h4>
```
Jackbk: Whitespace formatting

private-templates-service/client/src/Welcome.tsx
Jackbk: or just WelcomeContent(user: User) since I think that's all you are pulling from props

Also arrow functions

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { Container } from "reactstrap";
+import { UserAgentApplication } from "msal";
+import NavBar from "./NavBar";
+import ErrorMessage from "./ErrorMessage";
+import Welcome from "./Welcome";
+import config from "./Config";
+import { getUserDetails, getOrgDetails } from "./GraphService";
+import "bootstrap/dist/css/bootstrap.css";
+
+interface State {
```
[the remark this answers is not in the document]
kodyang: Left a todo, we will add it in a separate PR.

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,205 @@
...
+        cacheLocation: "localStorage",
+        storeAuthStateInCookie: true
+      }
+    });
+
+    var user = this.userAgentApplication.getAccount();
+
+    this.state = {
+      isAuthenticated: user !== null,
+      user: {},
```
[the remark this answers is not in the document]
kodyang: After a little digging, state.user is populated later on if user var exists. I will leave it as is.

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,205 @@
...
+            />{" "}
+          </Container>{" "}
+        </div>{" "}
+      </Router>
+    );
+  }
+
+  setErrorMessage(message: any, debug: any) {
+    this.setState({
+      error: {
```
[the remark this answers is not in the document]
kodyang: I found that this function was not used. deleted.

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <div>
+          <NavBar
+            isAuthenticated={this.state.isAuthenticated}
+            authButtonMethod={
+              this.state.isAuthenticated
+                ? this.logout.bind(this)
+                : this.login.bind(this)
+            }
+            user={this.state.user}
+          />{" "}
```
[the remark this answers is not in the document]
kodyang: We didn't update the linter config but the whitespace was not supposed to be there, removed.

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,205 @@
...
+    try {
+      await this.userAgentApplication.loginPopup({
+        scopes: config.scopes,
+        prompt: "select_account"
+      });
+      await this.getUserProfile();
+    } catch (err) {
+      var error = {};
+
+      if (typeof err === "string") {
```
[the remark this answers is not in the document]
kodyang: I changed it to a switch statement which I think is self documenting

private-templates-service/client/src/GraphService.tsx
kodyang: Added a todo, will be a part of adding redux. I think its outside the scope of this PR

private-templates-service/client/src/Welcome.tsx
kodyang: Will be a part of store.

private-templates-service/client/src/App.tsx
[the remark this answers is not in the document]
Jackbk: Oh, I see - you are just using this to check if the user exists before making further calls? 
``

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,212 @@
...
+  displayName?: string;
+  email?: string;
+  organization?: string;
+  avatar?: string;
+}
+
+// TODO: use react-redux
+interface State {
+  isAuthenticated: boolean;
+  user: UserType;
```
Jackbk: It is possible for a user to not be logged in right?  If so case we should make this able to be null or undefined. `user?: UserType`

private-templates-service/client/src/App.tsx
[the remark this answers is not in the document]
Jackbk: If the state allows the user to be null, we should set it null instead of empty here

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,212 @@
...
+import ErrorMessage from "./components/ErrorMessage/ErrorMessage";
+import Welcome from "./components/Welcome/Welcome";
+import config from "./Config";
+import { getUserDetails, getOrgDetails } from "./Services/GraphService";
+import { ErrorMessageProps } from "./components/ErrorMessage/ErrorMessage";
+
+import "bootstrap/dist/css/bootstrap.css";
+
+export interface UserType {
+  displayName?: string;
```
Jackbk: Is it possible for a user to be logged in, and not have these values? If not lets make them required so we don't need to do null checks and stuff later on

private-templates-service/client/src/App.tsx
kodyang: I think we populate user as `user: {}` and update it later on, i made that more explicit

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,212 @@
...
+        clientId: config.appId,
+        redirectUri: config.redirectUri
+      },
+      cache: {
+        cacheLocation: "localStorage",
+        storeAuthStateInCookie: true
+      }
+    });
+
+    var user = this.userAgentApplication.getAccount();
```
Jackbk: In typescript we normally use `let` instead of `var` as let is block scoped and var isnt: https://www.tutorialsteacher.com/typescript/typescript-variable

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  login = async () => {
+    try {
+      await this.userAgentApplication.loginPopup({
+        scopes: config.scopes,
+        prompt: "select_account"
+      });
+      await this.getUserProfile();
+    } catch (err) {
+      var error = {};
```
Jackbk: These var's should also be let

private-templates-service/client/src/App.tsx
```diff
@@ -0,0 +1,212 @@
...
+          user: {
+            displayName: user.displayName,
+            email: user.mail || user.userPrincipalName,
+            organization: org
+          },
+          error: null
+        });
+      }
+    } catch (err) {
+      var error = {};
```
Jackbk: If we type this from the start, it will ensure the code doesn't add extra props that aren't valid and also has any required ones (if you update the ErrorMessageProps type later for example, the compilation will break here - which is what you want).

Jackbk: `var error: ErrorMessageProps;`

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+function UserAvatar(props: { user: UserType }): ReactElement {
+  // If a user avatar is available, return an img tag with the pic
+  if (props.user.avatar) {
+    return (
+      <AvatarIcon
+        src={props.user.avatar}
+        alt="user"
+        className="rounded-circle align-self-center mr-2"
+        style={{ width: "32px" }}
```
Jackbk: Maybe I don't understand how your styling library works (I'll read up about it), but these styles should be in their own file somewhere

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+} from "reactstrap";
+
+// Components
+import { UserType } from '../../App';
+
+// Styles
+import "@fortawesome/fontawesome-free/css/all.css";
+import { AvatarIcon, DefaultAvatarIcon } from './styled';
+
+function UserAvatar(props: { user: UserType }): ReactElement {
```
Jackbk: This is kinda cool, I haven't seen this notation before.

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,127 @@
...
+  );
+}
+
+export interface NavBarProps {
+  isAuthenticated: boolean,
+  authButtonMethod: () => void,
+  user: UserType,
+}
+
+interface State {
```
Jackbk: NavBarState, since otherwise we will have two State interfaces (one from App.tsx and one here)

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: this was my mistake, it can be removed and styling would still be applied.

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: I updated the name.

I am not familiar with convention of how the types of states are defined in TS, though I thought it would be okay to have a `interface State {...}` in many files, as long as they are not exported. If they are exported, then they will be given a more unique name.

Jackbk: No you are right, as long as its not exported it is fine.  I was thinking props (in which the interface would be exported).  Resolving

private-templates-service/client/package.json
```diff
@@ -5,6 +5,13 @@
 	"dependencies": {
 		"@fortawesome/fontawesome-free": "^5.11.2",
 		"@microsoft/microsoft-graph-client": "^2.0.0",
+		"@types/jest": "^24.9.1",
```
turalf: It'd be better to put dependencies to dev dependencies as much as possible, to keep your bundle size smaller. 
Jest is not needed to run your code in prod.

private-templates-service/client/src/Services/GraphService.tsx
```diff
@@ -0,0 +1,30 @@
+
```
turalf: if you are not using jsx syntax .ts extension would be better for a file.

private-templates-service/client/src/Services/GraphService.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Client } from '@microsoft/microsoft-graph-client';
+import { AuthResponse } from 'msal';
+// var graph = require("@microsoft/microsoft-graph-client");
+
+function getAuthenticatedClient(accessToken: AuthResponse): Client {
+  // Initialize Graph client
+  const client = Client.init({
+    // Use the provided access token to authenticate
+    // requests
+    authProvider: (done: any) => {
```
Jackbk: done should have a type

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,125 @@
...
+  authButtonMethod: () => void,
+  user?: UserType,
+}
+
+interface State {
+  isOpen: boolean
+}
+
+export default class NavBar extends React.Component<NavBarProps, State> {
+  constructor(props: any) {
```
Jackbk: Props type

PR #19, opened 2020-01-25
Add authorization to backend

gracez72:
**Ticket Item:** 
https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/02?workitem=33261

- Update backend to use typescript
- Add authorization to the one endpoint


private-templates-service/package.json
```diff
@@ -5,12 +5,24 @@
 	"main": "server.js",
 	"scripts": {
 		"test": "echo \"Error: no test specified\" && exit 1",
-    "start:prod": "npm install && node server.js",
-    "start": "node server.js"
+		"start:prod": "npm install && node server.js",
+		"start": "tsc && node dist/server.js | bunyan"
```
kodyang: Previously it was using spaces but now the whitespace is tabs, can we keep using spaces?

private-templates-service/tsconfig.json
```diff
@@ -0,0 +1,67 @@
+{
```
kodyang: Can you document how this file was generated

private-templates-service/package.json
gracez72: Yep sounds good

PR #23, opened 2020-01-25
client side routing

kodyang:
small change that means when viewing server page, the route is properly served.


Jackbk:
I will review this once the other PR is in since a lot of it looks the same

PR #27, opened 2020-01-28
Add basic GET, POST template endpoints

gracez72:
Tickets: [TASK-33262:](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/02?workitem=33262), [TASK-33263](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/02?workitem=33263)

- get all templates
- get by template id
- post template - checks template is valid json, adds template owner as oid from access token


private-templates-service/src/controllers/template.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
+    bearer = bearer.split(/[ ]+/).pop();
+    if (!bearer) {
+        return res.status(400).send('Malformatted access token.');
+    }
+
+    var decodedTokenOid = jws.decode(bearer).payload.oid;
+    logger.info('Owner is: ' + decodedTokenOid);
+
+    const template = new Template({
```
Jackbk: Do you have a plan for extracting this logic to it's own package in future?  The sort of structure Matt was after (from what I understand) is that business logic like this will be in it's own package.  

The code here would then look something like:

import adaptivecarservice from 'adaptivecardsservice'
cardService = new adaptiveCardService(.......)

Then in places like this just:
cardService.saveCard(template)

This would allow other people to use all the business logic in their own web service if they want (and then the API here would just wrap the logic and provide authentication)

private-templates-service/src/controllers/template.ts
```diff
@@ -1,19 +1,78 @@
...
+    }
+
+    // Retrieve bearer token & decode oid
+    var bearer = req.get('Authorization');
+    if (!bearer) {
+        return res.status(400).send('No access token was found.');
+    }
+    bearer = bearer.split(/[ ]+/).pop();
+    if (!bearer) {
+        return res.status(400).send('Malformatted access token.');
```
Jackbk: nit: malformed

private-templates-service/src/controllers/template.ts
```diff
@@ -1,19 +1,78 @@
...
+    }), function(req: Request, res: Response, next: NextFunction){
     next();
 });
 
-// create a GET route
-templateRouter.get('/', (req : any, res : any) => {
-    res.send({
-        express: "YOUR EXPRESS BACKEND IS CONNECTED TO REACT. CONFIRMED"
+// GET templates route
+const getTemplates = (req: Request, res: Response, next: NextFunction) => {
```
Jackbk: Convention is to prefix unused variables with '_' -` _req`

private-templates-service/src/models/Template.ts
```diff
@@ -0,0 +1,14 @@
+import mongoose from 'mongoose';
+
+const templateSchema = new mongoose.Schema({
```
Jackbk: I think there should be another layer of abstraction here, as we don't want to lock the user to mongo

gracez72: Yep, we're working on a database adapter that will work for different databases - I've used a local mongo db for now just to set up the endpoints

private-templates-service/src/controllers/template.ts
gracez72: Fixed, thanks!

private-templates-service/src/controllers/template.ts
gracez72: Fixed!

private-templates-service/src/controllers/template.ts
gracez72: I'm not 100% sure how we would extract this logic right now, I'll look into it and follow up!

Jackbk: Cool, just something to keep in mind when writing this stuff

private-templates-service/src/app.ts
```diff
@@ -1,21 +1,52 @@
...
 app.use(express.static(path.join(__dirname, "./client/build")));
 app.use(passport.initialize());
 app.use(passport.session());
+app.use(
+  session({
+    resave: true,
+    saveUninitialized: true,
+    secret: "test",
+    store: new MongoStore({
+      url: "mongodb://localhost/adaptivecard",
```
Jackbk: You don't have to, but I would add a todo to move these parameters (specifically the URL) out to a config file somewhere so we dont forget later

gracez72: Done, thanks!

PR #28, opened 2020-01-28
Remove client dependency from backend

gracez72:
Added exclude field to tsconfig file


private-templates-service/tsconfig.json
```diff
@@ -1,68 +1,69 @@
...
+		// "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
 
-    /* Advanced Options */
-    "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
-  },
-  "lib": ["es2015"]
+		/* Advanced Options */
+		"forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
+	},
+	"exclude": ["./client"],
```
kodyang: i assume this is the only change? we will configure the linter to not change spaces to tabs.

Jackbk: Will ./client folder have it's own tsconfig?

gracez72: Yes, the client folder has its own tsconfig

PR #30, opened 2020-01-29
Add CODEOWNERS for PR review

markAtMicrosoft:
We will use this file to require a coach, sponsor or garage LT to review along with at least one intern each PR

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# At least one of the code owners below will be required on each PR:
+
+* @smmatte @rectified95 @rachel-slater @jackbk @turalf @markAtMicrosoft @matthidinger


.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# At least one of the code owners below will be required on each PR:
+
+* @smmatte @rectified95 @rachel-slater @jackbk @turalf @markAtMicrosoft @matthidinger
```
Jackbk: Will this affect the other people outside the internship working on this?  If so we can specify just the ./private-templates-service folder

markAtMicrosoft: The codeowners file is ONLY used when specifically assigned to a particular branch.  For example we will assign to the interns "dev" branch a requirement that at least one codeowner must approve a PR before it can merged into dev.

markAtMicrosoft: It will not affect any other branch or PR unless so designated.

PR #31, opened 2020-01-29
add concurrently and additional npm scripts

kodyang:
Our current dev environment is a little clunky, we need to run `npm run start` in two places and manually restart the server whenever we make a change.

Using `npm run dev`, we will now have both server and client running in a single terminal, and both toggled to auto-reload through webpack and nodemon.


private-templates-service/client/src/components/Welcome/Welcome.tsx
```diff
@@ -31,9 +31,11 @@ const WelcomeContent = (props: WelcomeProps): ReactElement => {
 
   // Not authenticated, present a sign in button
   return (
-    <Button color="primary" onClick={props.authButtonMethod}>
-      Click here to sign in
-		</Button>
+    <React.Fragment>
```
Jackbk: Since there is only one component, you don't need React.Fragment here.  (Fragment just allows you to return a group of components)

PR #32, opened 2020-01-29
Users/hagupt/instantiate dashboard routing

guptaharsh25:


private-templates-service/client/src/App.tsx
```diff
@@ -85,18 +86,21 @@ class App extends Component<{}, State> {
...
                   {...props}
                   isAuthenticated={this.state.isAuthenticated}
                   user={this.state.user}
                   authButtonMethod={this.login}
                 />
-              )}
-            />
+                )}
+              </Route>
+              <Route path="/dashboard">
```
kodyang: please add keyword exact `<Route exact path="/dashboard">`

guptaharsh25: done

private-templates-service/client/src/App.tsx
```diff
@@ -85,18 +86,21 @@ class App extends Component<{}, State> {
           />
           <Container>
             {error}
-            <Route
-              exact
-              path="/"
-              render={props => (
+            <Switch>
+              <Route exact path="/">
+                {this.state.isAuthenticated ? <Redirect to="/dashboard"/> : props => (
```
Jackbk: I would create an auth HOC and do this in the component itself

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+interface Props {}
+
+class Dashboard extends React.Component<Props, State> {
+	render() {
+		return <React.Fragment>DASHBOARD</React.Fragment>;
+	}
+}
+
+export default Dashboard;
```
Jackbk: As per my comment above, using a HOC would look something like:

`export default requireAuthentication(Dashboard)`

This will allow the logic to be easily changed in future

Example: https://riptutorial.com/reactjs/example/30238/higher-order-component-that-checks-for-authentication

private-templates-service/client/src/AuthenticatedComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface State {
+}
+interface AuthProps {
+  isAuthenticated: boolean,
+  user?: UserType,
+  authButtonMethod: () => Promise<void>
+}
+
+export function requireAuthentication(Component: any) {
```
Jackbk: This should be a generic, so that the component that uses it can pass its props too:

Something like:
`const  requireAuthentication = <P extends object>(Component: React.ComponentType<P>) => {

  return class AuthenticatedComponent extends React.Component<AuthProps & P> {`

private-templates-service/client/src/AuthenticatedComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Redirect } from "react-router-dom";
+import { UserType } from "./App";
+import Welcome from "./components/Welcome/Welcome";
+
+interface State {
+}
+interface AuthProps {
+  isAuthenticated: boolean,
+  user?: UserType,
+  authButtonMethod: () => Promise<void>
```
Jackbk: Does this need to pass all the way through? I think the <Welcome /> component itself should be able to handle this logic

private-templates-service/client/src/AuthenticatedComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function requireAuthentication(Component: any) {
+
+  return class AuthenticatedComponent extends React.Component<AuthProps, State> {
```
Jackbk: Since state is just an empty interface, you can just provide props - state isn't enforced

private-templates-service/client/src/AuthenticatedComponent.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
Jackbk: Should name the file the same as what is exported, in this case requireAuthentication

private-templates-service/client/src/AuthenticatedComponent.tsx
Jackbk: If you want me to explain this further let me know

private-templates-service/client/src/AuthenticatedComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+  isAuthenticated: boolean,
+  user?: UserType,
+  authButtonMethod: () => Promise<void>
+}
+
+export function requireAuthentication(Component: any) {
+
+  return class AuthenticatedComponent extends React.Component<AuthProps, State> {
+
+    isAuthenticated() {
```
Jackbk: I would get in the habit of using ES6 arrow notations for binding.  Also in this case you can simplify it even further:

`isAuthenticated = () => this.props.isAuthenticated`

private-templates-service/client/src/App.tsx
```diff
@@ -85,18 +86,19 @@ class App extends Component<{}, State> {
...
+              <Route exact path="/">
+                <Dashboard
                   isAuthenticated={this.state.isAuthenticated}
-                  user={this.state.user}
                   authButtonMethod={this.login}
                 />
-              )}
-            />
+              </Route>
+              {/* <Route exact path="/dashboard">
```
Jackbk: Can remove this now

private-templates-service/client/src/AuthenticatedComponent.tsx
Jackbk: Also, should be in a subfolder or something

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
```
Jackbk: You're welcome to define your own patterns for this - but the way I set up my components, the file structure looks like this:

/components/Dashboard
            - index.ts
            - Dashboard.ts
            - DashboardTests.ts

Then the Dashboard.ts would be this, and the index.ts file would be a super simple file like:
`import Welcome from "./Welcome";

export default Welcome `

Then when you use Welcome, you can import from "./components/Welcome" instead of "./components/Welcome/Welcome"

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { requireAuthentication } from "../../AuthenticatedComponent";
+
+interface State {}
+
+interface DashboardProps {
+	isAuthenticated: boolean
+	authButtonMethod: () => Promise<void>;
+}
+
+class Dashboard extends React.Component<DashboardProps, State> {
```
Jackbk: Can remove empty state

private-templates-service/client/src/AuthenticatedComponent.tsx
guptaharsh25: From my understanding the props are passed to the requireAuthentication function, which propagates the props to the Dashboard/Welcome component. If I remove AuthProps, I'm not sure how to pull the props from the Component itself

Jackbk: I meant specifically the authButtonMethod - can the authentication logic be handled by the class that owns the button itself?

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
guptaharsh25: Not sure if I understand this pattern. Could you elaborate?

Jackbk: Here is an example of a component library using this structure: https://github.com/pixeloven/typescript-react-component-library/tree/master/src/shared/components/atoms/Button

private-templates-service/client/src/AuthenticatedComponent.tsx
guptaharsh25: the login and logout functions were written in the App.tsx file, this authButtonMethod prop is used by Welcome and NavBar to define whether the user needs to log in/out. 

App.tsx: 
` <NavBar
            isAuthenticated={this.state.isAuthenticated}
            authButtonMethod={
              this.state.isAuthenticated
                ? this.logout
                : this.login
            }
            user={this.state.user}
   />`

If I take out the prop, wouldn't the login and logout function need to be migrated to the Welcome and NavBar components?

Jackbk: That raises a good point of where should that logic be haha.  The login stuff isn't specific to the App class, so it probably shouldn't be there anyway.  You could either extract the logic to a utils file which can be called from either place, or make use of something like redux saga (which is event driven like Redux, but for making async calls).  Setting up Saga now might make complex actions easier later: https://medium.com/@sprocompany/why-you-should-use-redux-saga-b64f6fce5e4a

private-templates-service/client/src/components/Welcome/Welcome.tsx
```diff
@@ -6,17 +6,18 @@ import { RootState } from "../../store/rootReducer";
...
 
-const mapStateToProps = (state: RootState) => {
-  return {
-    user: state.auth.user,
-  };
-}
-
 const WelcomeContent = (props: WelcomeProps): ReactElement => {
   // If authenticated, greet the user
   if (props.user && props.user.displayName) {
```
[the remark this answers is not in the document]
guptaharsh25: True, will remove the authenticated logic.

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+interface DashboardStates {
+	error: ErrorMessageProps | null;
+}
+
+interface DashboardProps {
+	isAuthenticated: boolean
+	user?: UserType
+	authButtonMethod: () => Promise<void>;
```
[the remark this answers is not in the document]
guptaharsh25: Will change it so that we use ; everywhere.

PR #33, opened 2020-01-29
Users/daniellemott/tsx backend testing

daniellemott:
#Task 33325: Configure Jest testing framework for Frontend and Backend
Configured frontend and backend tests. Pipelines were updated (@nnmkhang) so that all tests must pass before code is deployed. 

To run tests:
(from within private-templates-service)
npm install
npm run test


private-templates-service/server/dist/app.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
```
turalf: I guess you should ignore dist folder(or all generated files)

daniellemott: Just added dist/ to the .gitignore folder.

matthidinger: Good calling adding those to gitignore. If they were already commited into the repo you will have to `git rm` them to manually delete them; since gitignore only applies to new files

daniellemott: I just removed all files under dist/.

private-templates-service/client/package.json
```diff
@@ -56,7 +56,7 @@
 		"eslint-plugin-prettier": "^3.1.2",
 		"eslint-plugin-react": "^7.14.3",
 		"eslint-plugin-react-hooks": "^1.7.0",
-		"@types/jest": "^24.9.1"
+		"@types/jest": "^25.1.0"
```
kodyang: Does the version need to be updated?

private-templates-service/.gitignore
```diff
@@ -1,6 +1,7 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
-dist
+
```
Jackbk: nit: extra line

private-templates-service/client/src/__tests__/App.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
```
Jackbk: Are you going to have a _tests_ folder that has all the tests, or have each component have it's own test in its own folder?  I commented to Harsh yesterday a pattern I like, but I think it really needs to be defined now before the codebase gets too large 
Relevant discussion:
https://github.com/microsoft/adaptivecards-templates/pull/32#discussion_r372684998

private-templates-service/client/package.json
daniellemott: Changed back to 24.9.1

private-templates-service/server/package.json
```diff
@@ -0,0 +1,50 @@
...
+		"@testing-library/jest-dom": "^5.0.2",
+		"@testing-library/react": "^9.4.0",
+		"@types/async": "^3.0.5",
+		"@types/bluebird": "^3.5.29",
+		"@types/body-parser": "^1.17.1",
+		"@types/bunyan": "^1.8.6",
+		"@types/connect-mongo": "^3.0.0",
+		"@types/express-validator": "^3.0.0",
+		"@types/jws": "^3.2.1",
+		"@types/mongoose": "^5.5.43",
```
turalf: dev-dependencies would be better I guess? Those are needed only before build

private-templates-service/client/src/__tests__/App.test.tsx
daniellemott: I checked in with Harsh and we will place component test's with the component.

private-templates-service/server/package.json
daniellemott: I copy and pasted them into devDependencies.

PR #34, opened 2020-01-30
Redux

kodyang:
For now, store the user and an isAuthenticated bool into the store.


private-templates-service/client/src/store/auth/actions.ts
```diff
@@ -0,0 +1,16 @@
+import { LOGIN, LOGOUT, AuthTypes } from './types';
+
+export function login(user: any): AuthTypes {
```
nnmkhang: export function login(users: any): AuthTypes { } --> users: any can be replaced by users: userType

private-templates-service/client/src/components/NavBar/AuthNavItem.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { ReactElement } from "react";
```
kodyang: This function component was moved out of `NavBar.tsx`

private-templates-service/client/src/components/NavBar/UserAvatar.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactElement } from "react";
```
kodyang: This function component was moved out of NavBar.tsx

private-templates-service/client/src/components/Welcome/Welcome.tsx
```diff
@@ -37,9 +39,9 @@ const WelcomeContent = (props: WelcomeProps): ReactElement => {
       </Button>
     </React.Fragment>
   );
-}
+};
```
Jackbk: Don't need the ;
I wonder if your linter config is different from others in your team or something? These look auto added

private-templates-service/client/src/components/Welcome/Welcome.tsx
```diff
@@ -54,3 +56,7 @@ export default class Welcome extends React.Component<WelcomeProps> {
     );
   }
 }
+
+const VisibleWelcome = connect(mapStateToProps)(Welcome);
+
+export default VisibleWelcome;
```
Jackbk: Export/file name discrepancy

private-templates-service/client/src/store/auth/actions.ts
```diff
@@ -0,0 +1,16 @@
...
+    type: LOGIN,
+    text: 'User login',
+    user: user
+  }
+}
+
+export function logout(): AuthAction {
+  return {
+    type: LOGOUT,
+    text: 'User Logout'
```
Jackbk: Text on line 6 has a lowercase second letter and this one has uppercase, should change one of them so they are consistent

private-templates-service/client/src/App.tsx
```diff
@@ -201,12 +210,12 @@ class App extends Component<{}, State> {
...
       });
     }
-  }
+  };
 }
 
-export default App;
+const VisibleApp = connect(mapStateToProps, mapDispatchToProps)(App);
+
+export default VisibleApp;
```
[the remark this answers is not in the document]
kodyang: reading some docs shows they just do `export default connect(mapStateToProps, mapDispatchToProps)(App);` so I changed all the files to match this pattern.

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -1,90 +1,42 @@
...
+    isAuthenticated: state.auth.isAuthenticated,
+    user: state.auth.user
+  };
+};
 
 export interface NavBarProps {
-  isAuthenticated: boolean,
-  authButtonMethod: () => void,
-  user?: UserType,
+  authButtonMethod: () => void;
```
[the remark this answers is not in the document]
kodyang: I looked into this, it seems to be used interchangeably, typescript docs use semicolons but sites like tutorialpoint use commas. I will leave it as is since it seems like either will work the same.

private-templates-service/client/src/App.tsx
```diff
@@ -1,40 +1,59 @@
...
+// Components
 import NavBar from "./components/NavBar/NavBar";
 import ErrorMessage from "./components/ErrorMessage/ErrorMessage";
 import Welcome from "./components/Welcome/Welcome";
 import config from "./Config";
 import { getUserDetails, getOrgDetails } from "./Services/GraphService";
 import { ErrorMessageProps } from "./components/ErrorMessage/ErrorMessage";
 
+// CSS
 import "bootstrap/dist/css/bootstrap.css";
```
nnmkhang: Are we still using bootstrap?

kodyang: no but that is outside the scope of this PR

PR #37, opened 2020-02-04
Implement InMemoryDBAdapter

beervoley:
This pull request contains implementations of in memory db that uses Abstract Layer Interface for communication.
It still requires testing. I'm creating a pull request now so others can start incorporating ABL Interface in their code :)


private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
+import {Queries, Collections, StorageProvider, Interface} from './StorageProvider'
```
Jackbk: Tests?

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+            this.email = email;
+            if(_id) {
+                this._id = _id;
+            }
+        }
+
+        static fromJSON(j: string) : Collections.User | null{
+            try {
+                let user: Interface.IUser = JSON.parse(j);
+                return new User(user.team, user.org, user.email, user._id);
```
Jackbk: Can you just return user here?

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+
+module Collections {
+    export class User implements Interface.IUser {
+        _id?: string;
+        team: string[] = new Array();
+        org: string[] = new Array();
+        email: string;
+
+        
+        constructor(team: string | string[], org: string | string[], email:string, _id?: string,) {
```
Jackbk: I think you must have auto formatting not set up, maybe talk to @guptaharsh25 about how he set his up

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    export interface UserQuery {
+        userID? : string,
+        teamID? : string[],
+        orgID? : string[],
+        email?: string
+    }
+
+    // In case we need folders
```
Jackbk: Just remove this and add it in later if you want, no point having commented code cluttering it up

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
+// Base class for all the StorageProviders.
```
Jackbk: You should add a readme about the StorageProvider and how it is expected to be used

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+                    new Date(template.createdAt), new Date(template.updatedAt), template.isPublished, template._id);
+            } catch(e) {
+                logger.error(e);
+                return null;
+            }
+        }
+    }
+}
+
+abstract class StorageProvider {
```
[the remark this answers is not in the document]
Jackbk: It will also make testing easier to follow when you add tests

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+    //     parentID? : string,
+    //     childTemplatesID?: string[],
+    //     childFoldersID?: string[],
+    //     // permissions? : [enum]
+
+    // }
+}
+
+
+module Collections {
```
Jackbk: What is the purpose of Collections?  I think these are specific to InMemoryDBProvider?

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
+// Base class for all the StorageProviders.
+import logger from '../../util/logger'
+module Interface {
+
+    export interface IUser  {
```
Jackbk: Move to src/models ?

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+                // setTimeout(this._writeData, 1000); 
+            }
+        });
+    }
+
+    private async _readData(path:string, successMessage: string, options: Object) : Promise<string | void> {
+        return fs.readFile(path, options)
+        .then(buffer => {return buffer.toString()})
+        .catch(err => {
+            logger.error(sprintf("Cannot read from %s. Erros is %s.", path, err));
```
Jackbk: I think you can format directly through the logger: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bunyan/index.d.ts#L170

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
Jackbk: While this is a very full implementation of the provider, it could probably be as simple as this:

`module Interface {

    export interface IUser  {
        team: string[],
        org: string[],
        email: string
    }

    export interface ITemplateInstance {
        json: string,
        version: string
    }

    export interface ITemplate {
        instances: ITemplateInstance[],
        tags: string[],
        owner: string,
        createdAt: Date,
        updatedAt: Date,
        isPublished: boolean
    }
}


export class InMemoryDBProvider implements StorageProvider {

    users: Array<Interface.IUser> = [];
    templates: Array<Interface.ITemplate> = [];


    async  insertTemplate(template: Collections.Template) : Promise<void> {
        if(this.templates.includes(template)) {
            throw new Error("template already exists")
        }
        this.templates.push(template)
    }

    async insertUser(user: Collections.User) : Promise<void> {
        if(this.users.includes(user)) {
            throw new Error("user already exists")
        }
        this.users.push(user)
    }

    getTemplate(query: Queries.TemplateQuery) : Promise<Collections.Template[]> {
        let tmpTemplates = this.templates

        // logic to filter query

        return Promise.resolve(tmpTemplates)
    }


    getUser(query: Queries.UserQuery) : Promise<Collections.User[]> {
        let tmpUsers = this.users

        // logic to filter query

        return Promise.resolve(tmpUsers)
    }
}`

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+        return this._insert<Collections.User>(doc, this.users, this._commitUsers.bind(this), true);
+    }
+    async insertTemplate(doc: Collections.Template) : Promise<void> {
+        this._setTimestamps(doc);
+        return this._insert<Collections.Template>(doc, this.templates, this._commitTemplates.bind(this), true);
+    }
+    
+    async getTemplate(query: Queries.TemplateQuery): Promise<Collections.Template[]> {
+        let res: Collections.Template[] = new Array();
+        Array.from(this.templates.values()).forEach(template => {
```
Jackbk: Can be: `this.users.forEach`

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+            if(this._matchTemplate(query, template)) {
+                res.push(template)
+            }
+        })
+        return res;
+    }
+
+    async getUser(query: Queries.UserQuery): Promise<Collections.User[]> {
+        let res: Collections.User[] = new Array();
+        Array.from(this.users.values()).forEach(user => {
```
Jackbk: See comment about users

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+            if(this._matchUser(query, user)) {
+                res.push(user)
+            }
+        })
+        return res;
+        
+    }
+
+    getInstanceByVersion(template: Collections.Template, version:string) : Collections.TemplateInstance | undefined{
```
Jackbk: I dont think this is used anywhere

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+        return this._writeData(this.templatesPath, data, 
+                        sprintf(SUCCES_WRITE_MESSAGE, TEMPLATE_COLLECTION_NAME), options);
+    }
+
+    private _setTimestamps(doc: Collections.Template) : void{
+        doc.createdAt = new Date(Date.now());
+    }
+
+    async insertUser(doc: Collections.User) : Promise<void> {
+        return this._insert<Collections.User>(doc, this.users, this._commitUsers.bind(this), true);
```
Jackbk: I might be wrong about this, havent used TS in a while - but I think this will implicitly type if you just use the method:
`return this._insert(doc, this.templates, this._commitTemplates.bind(this), true);`

Jackbk: It will pull the generics out of the types you give it in params

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    }
+
+    private async _commitUsers() : Promise<void> {
+        let data: Array<Interface.IUser> = Array.from(this.users.values());
+        return this._commitCollection(this.usersPath,
+                        sprintf(SUCCES_WRITE_MESSAGE, USER_COLLECTION_NAME), data, options);
+    }
+
+    private async _commitTemplates() : Promise<void>{
```
Jackbk: If you use ES6 arrow function notations, these methods will automatically bind: _commitTempaltes = async (): Promise<void> => {

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
Jackbk: Actually what it looks like you have implemented is a a file storage database rather than in memory

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,175 @@
...
+            this.tags = this.tags.concat(tags);
+            this.owner = owner;
+            this.createdAt = createdAt;
+            this.updatedAt = updatedAt;
+            this.isPublished = isPublished;
+            if(_id) {
+                this._id = _id;
+            }
+        }
+        static fromJSON(j: string) : Collections.Template | null {
```
rachel-slater: For future reference:

A pattern I've seen alot & like - instead of returning the _explicit interface | null_ - you could create an additional interface called 'JSONResponse' (for example). It might look like:

```
export interface JSONResponse {
  success: boolean;
  errorMessage?: string;
  result?: any; // Or you could list the 3 possible classes here.
} 
```

Then this would be the return type. The pros are that the caller could read the 'success' property instead of being forced to check for null, they receive the errorMsg if anything goes wrong, and this interface is easily scalable if their are additional properties you want to include later.

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,169 @@
+import logger from "../../util/logger";
+
+module Interface {
+  export interface IUser {
```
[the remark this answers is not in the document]
gracez72: Nevermind, I spoke to Matt and we do need the user table!

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,206 @@
...
+    this._checkID(docToInsert);
+    if (!collection.has(docToInsert._id!)) {
+      collection.set(docToInsert._id!, docToInsert);
+    } else {
+      logger.error("Object with id: %s already exists. Insertion failed", doc._id!);
+    }
+  }
+
+  // Makes sure that id of the object is set
+  protected _checkID<T extends ITemplate | IUser>(doc: T) {
```
Jackbk: This should have a better name that describes what is actually happening, maybe 'upsertID'

private-templates-service/server/src/db/models/models.ts
```diff
@@ -0,0 +1,43 @@
+export interface IUser {
```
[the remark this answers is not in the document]
Jackbk: Yeah I think the storage provider ones need to go

private-templates-service/server/src/db/providers/StorageProvider.ts
```diff
@@ -0,0 +1,169 @@
...
+        return { success: true, result: new Collections.TemplateInstance(JSON.stringify(instanceData.json), instanceData.version) };
+      } catch (e) {
+        logger.error("Error with the JSON structure of the template instance. Object can not be parsed:\n" + e);
+        logger.error("Template Instance JSON: + " + j);
+        return { success: true, errorMessage: e };
+      }
+    }
+  }
+
+  export class Template implements Interface.ITemplate {
```
Jackbk: Are we using these classes now or can we get rid of them?

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+import logger from "../../util/logger";
+import uuidv4 from "uuid/v4";
+
+export class InMemoryDBProvider implements StorageProvider {
+  users: Map<string, IUser> = new Map();
+  templates: Map<string, ITemplate> = new Map();
+
+  constructor() {}
+
+  private _updateUser(user: IUser, updateQuery: UserQuery): void {
```
turalf: I feel like this function can be async as well.

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (updateQuery.orgID != undefined) {
+      user.org = updateQuery.orgID;
+    }
+    if (updateQuery.teamID != undefined) {
+      user.team = updateQuery.teamID;
+    }
+    this.users.set(user._id!, user);
+  }
+
+  private _updateTemplate(template: ITemplate, updateQuery: TemplateQuery): void {
```
turalf: consider async here as well

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+    doc.createdAt = new Date(Date.now());
+  }
+
+  protected _matchUser(query: UserQuery, user: IUser): boolean {
+    if (query.userID && !(query.userID === user._id)) {
+      return false;
+    }
+    if (query.email && !(query.email === user.email)) {
+      return false;
+    }
```
turalf: nit: you can combine those 2 checks into one with or statement

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (query.owner && !(query.owner === template.owner)) {
+      return false;
+    }
+    if (query.templateID && !(query.templateID === template._id)) {
+      return false;
+    }
+
+    if (query.isPublished && !(query.isPublished === template.isPublished)) {
+      return false;
+    }
```
turalf: nit: same as above - combine ifs

private-templates-service/server/src/db/providers/InMemoryDBProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (query.email && !(query.email === user.email)) {
+      return false;
+    }
+
+    if (query.orgID) {
+      for (let org of query.orgID) {
+        if (!user.org.includes(org)) {
+          return false;
+        }
+      }
```
turalf: it seems like there are a lot of repetition of this code.
You can write a function for that and call it in other places, as all the types are string and string[] I guess?

private-templates-service/server/src/db/models/models.ts
```diff
@@ -0,0 +1,43 @@
...
+  _id?: string;
+  team: string[];
+  org: string[];
```
turalf: shouldn't those names be plural? I might be missing something.

PR #39, opened 2020-02-04
Users/daniellemott/v2 recent template

daniellemott:
**Static Frontend that displays Adaptive Card**

Static frontend that renders Adaptive Cards in the dashboard. Will connect to backend and implement design revisions in future PRs. 

To view changes:
1. cd adaptivecards-templates\private-templates-service\client\src
2. **Change** redirectUri in Config.tsx (adaptivecards-templates\private-templates-service\client\src) to http://localhost:3000
3. npm install
4. npm start

### DIFF
--- a/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
+++ b/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
@@ -1,0 +1,20 @@
+import React from "react";
+import ReactDOM from "react-dom";
+
+import { Provider } from "react-redux";
+import { createStore } from "redux";
+import { rootReducer } from "../../store/rootReducer";
+import AdaptiveCard from "./AdaptiveCard";
+
+it("Renders without crashing", () => {
+	const div = document.createElement("div");
+	const store = createStore(rootReducer, {});
+	ReactDOM.render(
+		<Provider store={store}>
+			<AdaptiveCard />
+		</Provider>,
+		div
+	);
+
+	ReactDOM.unmountComponentAtNode(div);
+});
--- a/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
+++ b/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
@@ -7,14 +7,14 @@ import { rootReducer } from "../../store/rootReducer";
 import AdaptiveCard from "./AdaptiveCard";
 
 it("Renders without crashing", () => {
-	const div = document.createElement("div");
-	const store = createStore(rootReducer, {});
-	ReactDOM.render(
-		<Provider store={store}>
-			<AdaptiveCard />
-		</Provider>,
-		div
-	);
+  const div = document.createElement("div");
+  const store = createStore(rootReducer, {});
+  ReactDOM.render(
+    <Provider store={store}>
+      <AdaptiveCard />
+    </Provider>,
+    div
+  );
 
-	ReactDOM.unmountComponentAtNode(div);
+  ReactDOM.unmountComponentAtNode(div);
 });
--- a/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
+++ b/private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
@@ -3,6 +3,7 @@ import * as AdaptiveCards from "adaptivecards";
 import { Card, Container, TemplateName } from "../AdaptiveCard/styled";
 
 function getCard() {
+	// Hard coded, will remove and connect to backend in future PR
 	var card = {
 		type: "AdaptiveCard",
 		version: "1.0",
@@ -41,10 +42,6 @@ function GetRenderedCard() {
 		fontFamily: "Segoe UI, Helvetica Neue, sans-serif"
 	});
 
-	adaptiveCard.onExecuteAction = function(action) {
-		alert("Ow!");
-	};
-
 	var cardTemplate = getCard();
 	// Parse the card payload
 	adaptiveCard.parse(cardTemplate);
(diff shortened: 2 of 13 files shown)


private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import ReactDOM from "react-dom";
+
+import { Provider } from "react-redux";
+import { createStore } from "redux";
+import { rootReducer } from "../../store/rootReducer";
+import AdaptiveCard from "./AdaptiveCard";
+
+it("Renders without crashing", () => {
```
rachel-slater: Just FYI - thorough unit tests will check if the each function you've written achieves it's purpose. 
For example, in your case, you might test whether the 'GetCard()' function returns the expected data object. 

e2e tests will come in really handy for what you're doing here - you can automate all of the 'actions' - i.e. opening a browser, clicking on a button, checking if your rendered card HTML is on the page - which is probably the most accurate way of testing this.

But for now, I wouldn't change anything - I think it's a little too early to be worrying about robust unit tests.

daniellemott:
> can the tabs be changed to spaces?

done.

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from "react";
+import * as AdaptiveCards from "adaptivecards";
+import { Card, Container, TemplateName } from "../AdaptiveCard/styled";
+
+function getCard() {
+	// Hard coded, will remove and connect to backend in future PR
```
rachel-slater: FYI for the future, when you're removing the hard-coded data, you'll want to create an interface, which defines the structure of a card. For example:

```
interface CardData {
	type: boolean;
	version: string;
	body: any;
	actions: any;
}
```

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from "react";
+import * as AdaptiveCards from "adaptivecards";
+import { Card, Container, TemplateName } from "../AdaptiveCard/styled";
+
+function getCard() {
```
rachel-slater: As you're using Typescript, if the function returns something, you need to specify the _type_ of that return.

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+	// Hard coded, will remove and connect to backend in future PR
+	var card = {
```
rachel-slater: Just [FYI](https://stackoverflow.com/questions/35572895/what-is-the-difference-between-var-and-let-in-typescript).

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+				type: "Action.OpenUrl",
+				title: "GitHub",
+				url: "http://github.com/Microsoft/AdaptiveCards"
+			}
+		]
+	};
+	return card;
+}
+
+function GetRenderedCard() {
```
rachel-slater: I would split this function up into multiple smaller functions, for example:
- renderingSetup()
- parseCardTemplate() - you'd also need to handle an error case here - what if the parse function couldn't parse the data object you give it?
- renderAdaptiveCard()

private-templates-service/client/src/components/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+import AdaptiveCard from "../AdaptiveCard";
+import { Container } from "./styled";
+
+class Gallery extends React.Component {
+	render() {
+		return (
+			<Container>
+				<AdaptiveCard />
```
rachel-slater: This 3 x <.AdaptiveCard /.> is just a temp thing for now right?

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.test.tsx
daniellemott: Okay, that makes sense. I will add a ticket to our backlog to add more robust unit tests once we have a more base.

private-templates-service/client/src/components/Gallery/Gallery.tsx
daniellemott: Yes, I will "un-hardcode" it once I start grabbing adaptive card templates from the database (via the backend).

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
daniellemott: Done.

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
daniellemott: Do you have any suggestions on how to make an interface for the adaptive card if I don't know what's going to be in the card?

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
daniellemott: I wasn't too sure about what I was doing, but added some types. Also, for the function "renderAdaptiveCard()" on line 59, I didn't know how to create an HTMLElement so I left the return type as any. Would love some more feedback!

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
rachel-slater: Ohh, I thought this was the standard structure across all Cards! If the object doesn't have consistent keys, then it won't make sense to base it on an interface. So it's good as it is :)

private-templates-service/client/src/components/AdaptiveCard/AdaptiveCard.tsx
rachel-slater: You abstracted those into separate functions perfectly!! 👍 Much cleaner. And using 'any' is fine. (I had a quick search for a more appropriate HTML element type and couldn't find one either.)

PR #49, opened 2020-02-07
Users/vsevolod/mongo adapter

beervoley:
Implemented MongoDBAdapter that can be used to interact with mongo db. 
Insertion example if running locally:
image
For all the functions available look into `storageProviders/IStorageProvider`


private-templates-service/adaptivecards-templating-service/src/models/models.ts
```diff
@@ -1,27 +1,27 @@
 export interface IUser {
-    id?: string;
-    team: string[];
-    org: string[];
-    email: string;
-  }
-  
+	id?: string;
```
gracez72: Set up vscode to use spaces instead of tabs - can ask @guptaharsh25 about how to set it up

private-templates-service/adaptivecards-templating-service/src/models/mongo/TemplateModel.ts
```diff
@@ -0,0 +1,34 @@
...
+    versionKey: false,
+    timestamps: false
+  }
+);
+
+export const TemplateSchema: Schema = new Schema(
+  {
+    instances: { type: [TemplateInstanceSchema], required: true, default: [TemplateInstanceSchema] },
+    tags: { type: [String], default: [String] },
+    owner: { type: Schema.Types.String, required: true }, // add ref: "User" so it checks if owner exists and make type ObjectID
```
rachel-slater: Do these have to be 'Schema.Types' or can you just use 'String', 'Boolean' etc?

private-templates-service/adaptivecards-templating-service/src/models/mongo/UserModel.ts
```diff
@@ -0,0 +1,18 @@
+import mongoose, { Schema } from "mongoose";
+import { IUser } from "../models";
+
+export interface IUserModel extends mongoose.Document, IUser {
```
rachel-slater: Why not 'implement' the IUser interface, rather than extending it?

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,208 @@
...
+export class MongoDBProvider implements StorageProvider {
+  Template!: mongoose.Model<ITemplateModel>;
+  User!: mongoose.Model<IUserModel>;
+  db!: mongoose.Connection;
+  connectionString: string;
+  options: any;
+  async getUser(query: Partial<IUser>): Promise<JSONResponse<IUser[]>> {
+    return await this.User.find(query)
+      .then(users => {
+        if (users.length) {
```
rachel-slater: Is there supposed to be a condition here? ( .length > 0)?

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,208 @@
...
+          result: [],
+          errorMessage: "No templates found matching given criteria"
+        });
+      })
+      .catch(e => {
+        return Promise.resolve({ success: false, errorMessage: e });
+      });
+  }
+  // Updates Only one user
+  async updateUser(query: Partial<IUser>, updateQuery: Partial<IUser>): Promise<JSONResponse<Number>> {
```
rachel-slater: Simpler to use Boolean instead of Number? (Or if it's just used to indicate success, don't send a result, just use the 'success' flag).

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,208 @@
...
+        return Promise.resolve({
+          success: false,
+          number: 0,
+          errorMessage: e
+        });
+      });
+  }
+
+  // Set up mongoose models to work with the chosen collections
+  async setUpCollections(): Promise<JSONResponse<Boolean>> {
```
rachel-slater: I think this DB-specific stuff, from here, downwards - should be in a separate file.

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,208 @@
...
+      .close()
+      .then(result => {
+        return Promise.resolve({ success: true });
+      })
+      .catch(e => {
+        return Promise.resolve({ success: false, errorMessage: e });
+      });
+  }
+
+  constructor(connectionString: string, options: any) {
```
rachel-slater: When you've moved this to it's separate file, put the constructor at the top.

private-templates-service/adaptivecards-templating-service/src/models/mongo/TemplateModel.ts
beervoley: They can be ts types, you're correct
fixed

private-templates-service/adaptivecards-templating-service/src/models/mongo/UserModel.ts
beervoley: one interface cannot implement another, it has to extend it

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: success flag is gonna be sent, but I need to set the type of the **JSONResponse**, boolean would be redundant but with number I can just send 0 as # of updated users :)

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: but all of this code is mongo specific - it's implementing **IStorageProvider** as common interface, why would it be in a different location?:)

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: if length is 0, it will be evaluated to false anyway :)

rachel-slater: if you don't need to check length, then just `if (users)` ?

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
rachel-slater: - You've hard-coded the return numbers 0 and 1 (effectively the same as a boolean) so you're not providing any meaningful result information.
-The function name already indicates that one user will be updated, as it's called 'updateUser'.

I don't think you need to send a result in this case.

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: becayse if mongo doesn't find anything that matches the query - it will return emty array `[]` and for some weird reasons in typescript empty array passes if statement: `if([]) => true`
so I have to check for length :)

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: done :)

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
rachel-slater: There are two separate things going on in this file - functions used for database _operations_, and then the database _setup/management_ methods. The IStorageProvider interface specifies the db operations, not the setup.

From looking quickly, the _constructor_/_initializeDB_/_connect_/_close_ could all be easily moved to a new file (or a separate class within the same file if that's easier).

Tagging @Jackbk for second opinion.

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
rachel-slater: oh yeah, okay!

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
beervoley: You're right :)
Fixed :)
Though when you will look at `getUser|template` functions, you'll see that I will send the result in case of failure as `[]`. The reason for that is some weird type deduction by TS which is not working properly so it fails to compile. Screenshot attached :) 
image

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
Jackbk: I think @rachel-slater  has a good point here.  There is kinda two concepts going on. The MongoDBProvider can do adaptive card specific logic, e.g the `this.Template.findOneAndUpdate(query, updateQuery)` call. It should have a dependency on another layer which handles the grunt work.

Something like this stack:
AdaptiveCards Back end  -> MongoDBProvider -> MongoDBOperations

Anything not specific to the immediate running of adaptive cards (e.g, database set up) can be in operations.

It might seem a bit extra, but we have to remember that we are building an open source project and everything we commit from the start becomes an example for extensions later on.  If we can get a good pattern now, it will encourage any further DB providers to take a similar route.  And the easier we make it to understand the better/lower barrier to entry there will be.

beervoley: gotcha @rachel-slater , @Jackbk  - will do in a bit :)

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
rachel-slater: I think you can fix this by just specifying the type of 'users' (i.e. `.then((users: IUser[]) => ...` )?

I think for consistency we should decide now whether the 'result' of JSONResponse should be optional or not. I'm good with either, but we should make everything the same. If there is no meaningful info to return, and _result_ is optional, don't include it. In scenarios like this _updateUsers_ function, and if _result_ is non-optional, use Boolean.

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,195 @@
...
+      })
+      .catch(e => {
+        return Promise.resolve({
+          success: false,
+          errorMessage: e
+        });
+      });
+  }
+
+  async connect(): Promise<JSONResponse<Boolean>> {
```
rachel-slater: You can just have `connect `& `close` in the MongoWorker class right?

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -0,0 +1,198 @@
...
+    return userQuery;
+  }
+  private _constructTemplateQuery(query: Partial<ITemplate>): any {
+    let templateQuery: any = { ...query };
+    if (query.tags) {
+      templateQuery.tags = { $all: query.tags };
+    }
+    return templateQuery;
+  }
+  async getUsers(query: Partial<IUser>): Promise<JSONResponse<IUser[]>> {
```
gracez72: When no users exist, should getUsers return false or true? I think the MongoDBProvider returns false, but the InMemoryDBProvider returns true - might be useful to add comments to the StorageProvider so that all providers return consistent results and it's easier for the backend to use

beervoley: Yes, it needs to be fixed. I'll add the comments to the IStorageProvider to indicate expected behaviour.

PR #56, opened 2020-02-11
Add user endpoints & update template functionality to backend library

gracez72:
Ticket: [TASK-33551](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/04?workitem=33551)

**New changes to `adaptivecards-templating-service`:** 
- new user router
- posting a template automatically creates a new user object
- deleting a user removes both user and all templates created by the user
- insertUser and insertTemplate from `IStorageProvider.ts` returns the id of the created resource instead of 0/1
- user model has authID and issuer instead of email - authID represents the unique ID given by the access token and issuer represents an auth provider eg. AzureAD

**Existing endpoints on `server`:**
**GET** /user: returns only owner's information
**DELETE** /user: deletes user and all user's created templates
**GET** /template
       _isPublished_: If true, returns all published templates, regardless of owner eg. /template/?isPublished=true
**POST** /template/{id}: updates existing template 
**POST** /template: creates a new template

**Testing**
Follow the instructions [here](https://github.com/microsoft/adaptivecards-templates/tree/dev/private-templates-service#running-the-backend) for running the backend server and [here](https://github.com/microsoft/adaptivecards-templates/tree/dev/private-templates-service/adaptivecards-templating-service#developer-instructions) to link the backend library to the server.


private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -33,24 +34,168 @@ export class TemplateServiceClient {
...
+    
+    /**
+     * @public
+     * Deletes user info - can only delete own user and all templates created by that user
+     */
+    public async removeUser(): Promise<JSONResponse<Number>> {
+        let owner = this.authProvider.getOwner();
+        if (!owner){
+            return { success: false, errorMessage: ServiceErrorMessage.AuthFailureResponse };
+        }
```
turalf: This part can go into a method to remove code duplication

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -61,39 +206,77 @@ export class TemplateServiceClient {
...
      * @param {string} templateId - unique template id
      * @param {boolean} isPublished 
      * @param {string} templateName - name to query for
      * @param {string} version - version number
+     * @param {boolean} owned - If false, will retrieve all public templates regardless of owner
      * @returns Promise as valid json 
      */
-    public async getTemplates(templateId?: string, isPublished?: boolean, templateName?: string, version?: number): Promise<JSONResponse<ITemplate[]>> {
-        let owner = TemplateServiceClient.authProvider.getOwner();
+    public async getTemplates(templateId?: string, isPublished?: boolean, templateName?: string, version?: number, owned?: boolean): Promise<JSONResponse<ITemplate[]>> {
```
turalf: is templateId uniquely identifies template? If that is the  case the method name should be getTemplate?

turalf: templateId here is optional? what does it mean this to be optional? Do you need to specifically consider this case?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
gracez72: Done!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
gracez72: I added getTemplates to either get a template based on the id, or if an id was not passed (id is optional), the function can also retrieve all of the user's owner templates or published templates that are not owner by the user

PR #59, opened 2020-02-11
Updated build scripts and deployment pipelines

kodyang:
These were necessary for our updated project architecture, involving a `adaptivecards-templating-service` separate from the server

### DIFF
--- a/private-templates-service/server/package.json
+++ b/private-templates-service/server/package.json
@@ -6,9 +6,8 @@
   "scripts": {
     "test": "jest",
     "test:watch": "jest --watch",
-    "start:prod": "npm install && node server.js",
-    "start": "tsc && node dist/server.js | bunyan",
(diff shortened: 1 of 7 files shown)


private-templates-service/server/package.json
```diff
@@ -24,7 +24,10 @@
     "express-validator": "^6.3.1",
     "jws": "^4.0.0",
     "passport": "^0.4.1",
-    "passport-azure-ad": "^4.2.1"
+    "passport-azure-ad": "^4.2.1",
+    "axios": "^0.19.2",
+    "mongoose": "^5.8.10",
+    "uuid": "^3.4.0"
```
kodyang: @grace These seem required for it to build properly. Even though I am linking `adaptivecards-templating-service` and also requiring it above, its dependency's are not downloaded.

PR #63, opened 2020-02-12
Update deployment pipelines to accommodate new architecture

kodyang:
Not much to say here, this just makes things work.

## Fixes:
1. Test failure regarding `monaco-editor` was mocked
2. Test failure regarding 'useHistory' in sidebar was fixed by adding routing to test
3. Deployment works for real
4. prettier removed
5. Package lock cleaned up


private-templates-service/adaptivecards-templating-service/index.ts
```diff
@@ -0,0 +1,9 @@
+import { TemplateServiceClient } from '../adaptivecards-templating-service/src/TemplateServiceClient'
```
guptaharsh25: We can remove this file. Grace has made docs on how to make the import in `server/src/app.ts` work over here: https://github.com/microsoft/adaptivecards-templates/tree/dev/private-templates-service/adaptivecards-templating-service#developer-instructions

private-templates-service/server/src/app.ts
```diff
@@ -4,20 +4,20 @@ import passport from "./config/passport";
 import bodyParser from "body-parser";
 
 // import controllers
-import { TemplateServiceClient, ClientOptions, AzureADProvider, InMemoryDBProvider } from 'adaptivecards-templating-service';
+import { TemplateServiceClient, ClientOptions, AzureADProvider, InMemoryDBProvider } from '../../adaptivecards-templating-service';
```
guptaharsh25: This change can be reverted since the import is described here: https://github.com/microsoft/adaptivecards-templates/tree/dev/private-templates-service/adaptivecards-templating-service#developer-instructions

PR #68, opened 2020-02-12
Users/vsevolod/filter monggo

beervoley:
- Fixed duplicate ID generation when adding documents in a short time interval one by one in Mongo
- Implemented sort of templates by **name, dateCreated, dateModified**

Usage of sorting:
`let db: StorageProvider = new MongoDBProvider({connectionString: "..."}"`
`await db.connect()` 
`db.getTemplates({}, SortBy.name, SortOrder.ascending)`


private-templates-service/adaptivecards-templating-service/src/models/models.ts
```diff
@@ -37,6 +37,17 @@ export interface JSONResponse<T> {
   result?: T;
 }
 
+export enum SortBy {
```
rachel-slater: Instead of specifying _enum_ values, could it just be:

```
export enum SortBy {
  dateCreated,
  dateModified,
  name
}
```
or are the string values needed by mongo?

private-templates-service/adaptivecards-templating-service/src/models/mongo/TemplateModel.ts
```diff
@@ -33,3 +38,11 @@ export const TemplateSchema: Schema = new Schema(
     timestamps: { createdAt: true, updatedAt: true }
   }
 );
+TemplateSchema.pre("validate", function(next) {
+  this._id = MongoUtils.generateUniqueID().toHexString();
+  next();
+});
+
+TemplateSchema.index({ createdAt: 1 });
```
rachel-slater: Don't quite understand these 3 lines - guessing they'll be changed later?

private-templates-service/adaptivecards-templating-service/src/models/models.ts
beervoley: yup, they're essentially keys which you search by:)

private-templates-service/adaptivecards-templating-service/src/models/mongo/TemplateModel.ts
beervoley: These 3 lines create index for the keys which we will run sort on.

PR #72, opened 2020-02-14
Template Preview Page and Publish

kodyang:
[AB#33631](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33631)
[AB#33485](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33485)

# Description
Publishing is the feature that enables a template to be retrieved by developers through the API endpoint. Before a template is published, it is considered to be a "draft", which is only available to the creator and not accessible through the endpoint. We expect most devs will use the API in order to get templates to render on their application. 

This PR is mostly frontend UI, with some code to update the template to a published state leveraging the client SDK.

# Screenshots
## Template preview page
image

## Publish Modal 
image

## Template preview page of published template (look at green dot vs orange)
image

# Known bugs
- User can attempt to publish an already published template. We will leave this in for now. It will be fixed implicitly once we do unpublish, which will replace the `publish` button.
- Hitting publish causes the modal to disappear momentarily. This will be fixed when we add the loading/spinner component.


private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -101,4 +133,32 @@ export function updateTemplate(templateID: string, templateJSON: string, sampleD
...
+              templateObject,
+              templateObject.instances[0].json,
+              templateObject.name,
+              ""
+            )
+          )
+        }
+        dispatch(requestTemplateFailure());
+      })
+    } catch {
```
guptaharsh25: catch in new line

private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -69,15 +75,41 @@ function failureExistingTemplateUpdate(error: IncomingMessage): CurrentTemplateA
...
 
     let newTemplate = new TemplateJSON();
     newTemplate.template = templateJSON;
     newTemplate.name = templateName;
-    newTemplate.isPublished = false;
+    newTemplate.isPublished = isPublished || false;
```
guptaharsh25: can't this just be newTemplate.isPublished = isPublished ?

private-templates-service/client/src/store/currentTemplate/reducers.ts
```diff
@@ -64,7 +69,32 @@ export function currentTemplateReducer(state = initialState, action: CurrentTemp
...
+        sampleDataJSON: undefined,
+      }
+    case GET_TEMPLATE_SUCCESS:
+      return {
+        ...state,
+        template: action.template,
+        templateJSON: action.templateJSON,
+        templateName: action.templateName,
+        sampleDataJSON: action.sampleDataJSON,
+        isFetching: false,
```
guptaharsh25: for better readability, can state be assigned in the same order for each action type case

private-templates-service/client/src/store/currentTemplate/types.ts
```diff
@@ -1,10 +1,13 @@
...
 export interface CurrentTemplateState {
-  templateID?: string
-  templateJSON?: string
-  templateName?: string
-  sampleDataJSON?: string
+  templateID?: string;
+  templateJSON?: string;
+  templateName?: string;
+  sampleDataJSON?: string;
+  template?: Template;
```
guptaharsh25: Instead of having a separate templateID, templateJSON, templateName and sampleDataJSON field in state, now that we have a Template type state here can we change the store so that this state is only concerned with `template`?

Just a suggestion.

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -42,29 +56,37 @@ class Dashboard extends React.Component<Props, State> {
...
+  }
+
+  selectTemplate = (templateID: string) => {
+    this.props.getTemplate(templateID);
+    this.setState({ isPreviewOpen: true });
   }
 
   toggleModal = () => {
-    this.setState({ isPreviewOpen: !this.state.isPreviewOpen });
+    this.setState({ isPreviewOpen: false });
```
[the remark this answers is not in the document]
kodyang: I changed it back

private-templates-service/client/src/store/currentTemplate/actions.ts
kodyang: newTemplate.isPublished is of type `boolean | undefined` so if its undefined it should be assigned false

private-templates-service/client/src/store/currentTemplate/types.ts
kodyang: Can we do that in a separate pr? I think that is not a trivial change.

private-templates-service/client/src/store/currentTemplate/actions.ts
guptaharsh25: oh okay makes sense

private-templates-service/client/src/store/currentTemplate/types.ts
guptaharsh25: Sure.

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -51,29 +61,35 @@ class Dashboard extends React.Component<Props, State> {
...
         </DashboardContainer>
       );
     }
     //TODO add sort functionality to separate templates displayed in recent vs draft
-    let templates = undefined;
-    if (!this.props.batchTemplates.isFetching && this.props.batchTemplates.templateList) {
-      templates = this.props.batchTemplates.templateList.templates;
+    let templates = new Array<Template>();
+    if (!this.props.templates.isFetching
+      && this.props.templates.templates
```
turalf: This part is a little fuzzy for me. Why you have multiple levels of the same type? or is it a naming issue?

kodyang: yeah, really unfortunate naming here.

`this.props.templates` is what I named the variable.
`this.props.templates.templates` is a `TemplateList` object, containing a list as well as some additional data
`this.props.templates.templates.templates` is a 'Templates[]` object.

PR #75, opened 2020-02-19
Try / Catch block to handle parsing error

kodyang:
Thanks Danielle for the fix


private-templates-service/client/src/components/Common/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -55,9 +55,27 @@ function cleanTemplate(temp: TemplateInstance): Template {
   const templateString = JSON.stringify(temp.json);
   const replaceChar = templateString.replace(/\\\\\\/g, '');
   const trimTemp = replaceChar.slice(3, replaceChar.length - 3);
-  const jsonTemp = JSON.parse(trimTemp);
-  return jsonTemp;
 
+  let jsonTemp = JSON.parse("{}");
```
rachel-slater: Think you can just declare it as: `let jsonTemp = {};`

PR #86, opened 2020-02-24
Users/daniellemott/deploy to web

daniellemott:
**Task 33785**

Added final steps necessary to transition from inMemory to MongoDB. Added unit tests in private-templates-service/adaptivecards-templating-service to the pipeline and updated dev/stage pipelines to run these tests before deploying Azure Web App

_What I did to test this_
I ran our web app locally. I made a new template via the designer, saved it, and ensured it appeared on the dashboard.


private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -86,6 +86,7 @@ export class MongoDBProvider implements StorageProvider {
   // Updates Only one user
   async updateUser(query: Partial<IUser>, updateQuery: Partial<IUser>): Promise<JSONResponse<Number>> {
     let userQuery: any = this._constructUserQuery(query);
+    updateQuery = MongoUtils.removeUndefinedFields(updateQuery);
```
beervoley: This change as well as the one below is to address results of the fields updated that are not defined in the updateQuery. They were set to null by Mongo.

daniellemott: Thanks Seva.

PR #100, opened 2020-02-27
Fix Publish, Timestamp, and Card Formatting

kodyang:
## Description
This short PR fixes three things:
1. Previously, publish was not working because the definition of template.json changed from a string to a JSON object. The fix was to stringify before passing into the function
2. Date timestamp was previously ISO time, its now parsed to a more human readable time
3. Cards are formatted to have some additional spacing, minor change.


private-templates-service/client/src/components/Dashboard/PreviewModal/TemplateInfo/TemplateInfo.tsx
```diff
@@ -105,30 +105,37 @@ class TemplateInfo extends React.Component<Props, State> {
...
       tags,
-      createdAt, 
+      createdAt,
       instances,
     } = this.props.template;
     const { onClose } = this.props;
 
+    let createdAtParsed = "";
+
+    if (createdAt) {
```
rectified95: personal preference not to make if checks against non-Boolean types, it's a nit, though

PR #113, opened 2020-03-02
Update client to use client flag to track template hits

gracez72:
This PR makes sure the client accessing template by id does not increase the number of hits on the template. 

**Changes**
- Fix mongo provider bug using hits instead of numHits field 
- Add client query param flag to client SDK definition + update client
- Fix client SDK data type to any (for JSON object) instead of string

### DIFF
--- a/private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
+++ b/private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
@@ -1003,7 +1003,8 @@ export class TemplateServiceClient {
 
       let isPublished: boolean | undefined = req.query.isPublished ? req.query.isPublished === "true" || req.query.isPublished === "True" : undefined;
       let owned: boolean | undefined = req.query.owned ? req.query.owned === "true" || req.query.owned === "True" : undefined;
-      let isClient: boolean = req.body.isClient;
+      let isClient: boolean | undefined = req.query.isClient ? req.query.isClient === "true" || req.query.isClient === "True" : undefined;
+      
       if (!req.is("application/json")) {
         isClient = req.body.isClient === "true" || req.body.isClient === "True";
       }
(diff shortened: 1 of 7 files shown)


private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1003,7 +1003,8 @@ export class TemplateServiceClient {
 
       let isPublished: boolean | undefined = req.query.isPublished ? req.query.isPublished === "true" || req.query.isPublished === "True" : undefined;
       let owned: boolean | undefined = req.query.owned ? req.query.owned === "true" || req.query.owned === "True" : undefined;
-      let isClient: boolean = req.body.isClient;
+      let isClient: boolean | undefined = req.query.isClient ? req.query.isClient === "true" || req.query.isClient === "True" : undefined;
+      
       if (!req.is("application/json")) {
         isClient = req.body.isClient === "true" || req.body.isClient === "True";
```
markAtMicrosoft: would it not be better to have "req.query.isClient.toLowerCase( ) === "true" instead of doing everything twice?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1061,10 +1062,7 @@ export class TemplateServiceClient {
     });
 
     router.get("/:id?", (req: Request, res: Response, _next: NextFunction) => {
-      let isClient: boolean = req.body.isClient;
-      if (!req.is("application/json")) {
-        isClient = req.body.isClient === "true" || req.body.isClient === "True";
-      }
+      let isClient: boolean | undefined = req.query.isClient ? req.query.isClient === "true" || req.query.isClient === "True" : undefined;
```
markAtMicrosoft: toLowerCase( )

gracez72: Fixed!

PR #114, opened 2020-03-02
 fixing search bug

nnmkhang:
Adding quick fix for a search page bug.

The bug:
templates on the search page were not being properly rendered when clicked and would only show the previously rendered card that was clicked on the dashboard. 

The Solution:
This bug was caused since the seachpage did not properly call the getTemplate action.


kodyang:
Merged on Khang's request

PR #119, opened 2020-03-03
Create modal HOC to position modal closer to app root

kodyang:
[AB#33910](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33910)

## Bug Description
Previously, users can interact with the navbar even when a modal is open. This is because the modal is previously a child of a dashboard component. No matter how high the z-index is set, it will never be positioned higher than the navbar, which only shares a common ancestor node. 

## Solution
The modal must be positioned higher up in the DOM so the navbar and the modal are direct decedents of each other. This was done easily with [React Portals](https://reactjs.org/docs/portals.html). From now on, all modals should use the ModalHOC component.

### DIFF
--- a/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
+++ b/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 
 // Libraries
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
-import { PrimaryButton, ThemeSettingName } from 'office-ui-fabric-react'
+import { PrimaryButton } from 'office-ui-fabric-react'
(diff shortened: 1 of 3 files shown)


private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
```diff
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 
 // Libraries
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
-import { PrimaryButton, ThemeSettingName } from 'office-ui-fabric-react'
+import { PrimaryButton } from 'office-ui-fabric-react'
```
nnmkhang: nit: missing ;

PR #120, opened 2020-03-03
Replace hardcoded access token in client

gracez72:
This PR replaces the hardcoded access token for the one returned by AAD. 

**Changes**
- To run backend, run `npm start dev` or `npm start staging` or `npm start` for localhost (the env parameter changes the app ID)
- Backend library now checks the audience field of the token to see if it matches the app ID mentioned earlier
- Client passes logged in user access token to backend instead of hardcoded value


private-templates-service/adaptivecards-templating-service/src/authproviders/Config.ts
```diff
@@ -0,0 +1,7 @@
+export default {
+  appId: process.argv[2] === 'dev' ?
+    "4803f66a-136d-4155-a51e-6d98400d5506" : 
+    process.argv[2] === "staging" ? 
+    "ed0125aa-a0c3-48d5-8e2f-cd05d858e7ef" : 
+    "cc02a0f4-ef1b-4513-a431-9aca7b2f7fca"
```
kodyang: Where are the corresponding changes to the startup command? How is the `argv` populated exactly.. I would have expected a change in the package.json or in the build pipelines

gracez72: To run the backend, we changed the web app start up command to `npm start dev` (dev is the value retrieved by process.argv[2])

kodyang: makes sense, thanks !

PR #121, opened 2020-03-03
Hotfix for npm link-client 

nnmkhang:
Fixing versioning bug. 

### The Bug
When adding a tag, a new version would be created 

### The Reason
The reason behind this bug was because the frontend was not passing a version to the Client SDK. In the backend, if there is no version specified the version is auto incremented. This resulted in a new version each time a new tag was added.

### The Fix
To fix, `version` was added as a param into all instances of the updateTemplate call in the frontend.  

### Added hotfix
There was a small bug involving the npm run link-client script with the sudo command. I have resolved the bug by removing sudo and updated the docs to reflect the steps for running the application.

### DIFF
--- a/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
+++ b/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 
 // Libraries
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
-import { PrimaryButton, ThemeSettingName } from 'office-ui-fabric-react'
+import { PrimaryButton } from 'office-ui-fabric-react';
 import { SearchBox } from 'office-ui-fabric-react';
 
 import { Template, PostedTemplate } from 'adaptive-templating-service-typescript-node';
@@ -13,6 +13,7 @@ import { updateTemplate } from '../../../store/currentTemplate/actions';
 
 // Components
 import AdaptiveCard from '../AdaptiveCard';
+import ModalHOC from '../../../utils/ModalHOC';
 
 // Styles
 import {
@@ -92,4 +93,4 @@ class PublishModal extends React.Component<Props> {
   }
 }
 
-export default connect(() => { return {} }, mapDispatchToProps)(PublishModal);
+export default ModalHOC(connect(() => { return {} }, mapDispatchToProps)(PublishModal));
--- a/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
+++ b/private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
@@ -36,13 +36,13 @@ interface Props {
   template: Template;
   templateVersion: string;
   toggleModal: () => void;
-  publishTemplate: () => void;
+  publishTemplate: (templateVersion: string) => void;
 }
 
 const mapDispatchToProps = (dispatch: any) => {
   return {
-    publishTemplate: () => {
-      dispatch(updateTemplate(undefined, undefined, undefined, undefined, PostedTemplate.StateEnum.Live));
+    publishTemplate: (templateVersion: string) => {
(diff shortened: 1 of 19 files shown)


private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -110,6 +110,11 @@ export function updateTemplate(templateID?: string, templateJSON?: string, sampl
 
     let newTemplate = new PostedTemplate();
     const id = templateID || appState.currentTemplate.templateID;
+    
+    const version = currentVersion ? currentVersion :
```
gracez72: Should this default update the latest version instead of the oldest version?

nnmkhang: good catch

private-templates-service/client/src/components/Common/PublishModal/PublishModal.tsx
```diff
@@ -51,7 +51,7 @@ class PublishModal extends React.Component<Props> {
 
   publish = () => {
     // TODO: kodyang, grzhang PUBLISH CURRENT VERSION, NOT 1ST VERSION
```
kodyang: Can you remove the TODO comment?

private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -110,7 +110,11 @@ export function updateTemplate(templateID?: string, templateJSON?: string, sampl
 
     let newTemplate = new PostedTemplate();
     const id = templateID || appState.currentTemplate.templateID;
-
+    
+    const version = currentVersion || 
+    (appState.currentTemplate.template && appState.currentTemplate.template.instances && appState.currentTemplate.template.instances.length > 0 ?
+    appState.currentTemplate.template!.instances![0].version : "1.0")
```
kodyang: firstly, are the `!` needed in this context? 

secondly, what is the expected version from `appState.currentTemplate.template!.instances![0].version`? Is that the latest or the oldest available version?

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -117,7 +117,7 @@ function initDesigner(): ACDesigner.CardDesigner {
 
 function onSave(designer: ACDesigner.CardDesigner, props: DesignerProps): void {
   if (props.templateJSON !== JSON.stringify(designer.getCard()) || props.sampleDataJSON !== designer.sampleData) {
-    props.updateTemplate(props.templateID, JSON.stringify(designer.getCard()), props.templateName, designer.sampleData);
+    props.updateTemplate(props.templateID, "1.0", JSON.stringify(designer.getCard()), props.templateName, designer.sampleData); // TODO: get the version dynamically
```
kodyang: Is this going to be added by whoever is doing `edit`? Could you add `guptaharsh25` to the comment if so

private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -159,7 +159,7 @@ export function getTemplate(templateID: string) {
...
 
     try {
-      api.templateById(templateID, undefined, true).then((resp: any) => {
+      api.templateById(templateID, undefined, undefined,true).then((resp: any) => {
```
rachel-slater: Not sure why this has been changed.
Isn't this the API signature for _templateById_?
`public templateById (templateId: string, isPublished?: boolean, isClient?: boolean, version?: string, options: any = {})`

nnmkhang: There was an issue on with merging and building that caused this to happen, I have reverted the change and it should be good now

nnmkhang:
Hi @rachel-slater  This PR seems to not fix the bugs because a lot of the bugs that I was working on was fixed in @gracez72 's PR. After I merged dev the changes the files have changed. If you want to see my specific changes for the bugs please look at my first commit.

PR #129, opened 2020-03-05
Fix render bug

kodyang:
[AB#33964](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33964)

## Description
For unknown reasons, a small number of templates do not render properly on the template preview page. This one liner fixes that.

### Previously:
image

### Now: 
image


private-templates-service/client/src/components/Dashboard/PreviewModal/styled.tsx
```diff
@@ -36,6 +36,7 @@ export const ACPanel = styled.div`
 
 export const ACWrapper = styled.div`
   padding-bottom: 160px;
+  max-width: 100%;
```
nnmkhang: great work

PR #131, opened 2020-03-05
Users/nnmkhang/fix muti tags

nnmkhang:
[AB#33862](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33862)
Overview of This PR: 
- Fixed bug of multiple tags
- Added animation if user tries to add a tag that has already been created
- Added closing functionality (If user presses esc, clicks outside of the input form or tries to submit an empty tag, the tag will close)


private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -16,6 +16,7 @@ import {
   TagSubmitButton,
   TagSubmitIcon,
 } from './styled';
+import { runInThisContext } from 'vm';
```
kodyang: remove

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -45,15 +46,21 @@ interface State {
   newTagName: string;
 }
 
+interface RefsDict {
```
rachel-slater: Is this interface used anywhere else? If not, I don't think it makes sense to have it - it overcomplicates things imo.

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -81,12 +83,45 @@ class Tags extends React.Component<Props, State>  {
...
     }
   }
 
   closeAddTag = () => {
     this.setState({ isAdding: false });
+    this.setState({newTagName: ""});
+  }
+
+  onKeyDown = (keyStroke: any) => {
+    if(keyStroke.keyCode === 27){
```
turalf: I guess it would be better to have some kind of enum to keep keyCodes. It would be more readable, also in the feature if you add more keyCodes the code will still be more readable

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -6,6 +6,8 @@ import { RootState } from '../../../store/rootReducer';
 
 import { Template } from 'adaptive-templating-service-typescript-node';
 
+import Key  from '../../../globalKeyCodes';
```
kodyang: remove extra whitespace between `Key` and `from`

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -81,12 +85,43 @@ class Tags extends React.Component<Props, State>  {
...
     if (this.addTagInput && this.addTagInput.current && this.props.template && this.props.template.tags) {
       const tag = this.addTagInput.current.value;
-      this.props.updateTags([...this.props.template.tags, tag]);
+      if(!this.props.template.tags.includes(tag)){
+        if(tag === ""){
+          this.closeAddTag();
+        }
+        else{
+          this.props.updateTags([...this.props.template.tags, tag]);
+       }
```
kodyang: how does this line only have 7 spaces ?

kodyang: it is hard to tell which bracket it corresponds to

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -81,12 +85,43 @@ class Tags extends React.Component<Props, State>  {
     e.preventDefault();
     if (this.addTagInput && this.addTagInput.current && this.props.template && this.props.template.tags) {
       const tag = this.addTagInput.current.value;
-      this.props.updateTags([...this.props.template.tags, tag]);
+      if(!this.props.template.tags.includes(tag)){
```
kodyang: Rather than doing 
`if (! ... ) {
}
else {
}`

Can we do 
`if (...) {
} else {}`

private-templates-service/client/src/components/Common/Tags/Tags.tsx
```diff
@@ -81,12 +85,43 @@ class Tags extends React.Component<Props, State>  {
...
+        if(tag === ""){
+          this.closeAddTag();
+        }
+        else{
+          this.props.updateTags([...this.props.template.tags, tag]);
+       }
+      }
+      else{
+        this.highlightTag(tag,this.props.template.tags)
```
kodyang: space after `,` and add `;` eol

private-templates-service/client/src/globalKeyCodes.ts
```diff
@@ -0,0 +1,6 @@
+enum Key {
+  ESC = 27,
+  END = 35
+}
```
kodyang: Can we rename this to `KeyCode`

PR #140, opened 2020-03-11
Remove sensitive information from models and storage

beervoley:
In this PR I addressed [ACB33917](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/33917) and removed HCI from our models and database.
We previously stored first name, last name, team and organization of each user. We don't do it anymore. The only piece of information that's HCI classified is oid (authId in user model). This will be addressed in future PR.


kodyang:
Do any frontend changes need to be made?

beervoley:
> Do any frontend changes need to be made?

Not at the moment, no

private-templates-service/adaptivecards-templating-client/swagger.yaml
```diff
@@ -517,18 +505,8 @@ definitions:
...
-        type: array
-        items: 
-          type: string
-      org: 
-        type: array
-        items: 
-          type: string
   TemplateInfo: 
     type: object
     properties:
```
gracez72: Can remove the PostedUser definition and post user endpoint as well!

kodyang: has this been resolved?

PR #141, opened 2020-03-11
Users/hagupt/publish bug

guptaharsh25:
fixes multiple publish related bugs:

- sometimes backend would give 400 bad request when you hit publish, that is resolved
- if backend does give a 400, we're catching it so that the front-end doesn't hang
- user can now publish from deprecated state, which creates a new published version of the template and maintains the deprecated version
- instead of reading state of template instance from the template, we now read it from the template instance so that deprecated state shows up properly.

### DIFF
--- a/private-templates-service/client/src/store/currentTemplate/actions.ts
+++ b/private-templates-service/client/src/store/currentTemplate/actions.ts
@@ -17,10 +17,7 @@ import {
 
 import { Template, TemplateApi, PostedTemplate } from "adaptive-templating-service-typescript-node";
 
-import { IncomingMessage } from "http";
 import { RootState } from '../rootReducer';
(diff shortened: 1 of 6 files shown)


private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -197,8 +193,10 @@ export function updateTemplate(templateID?: string, currentVersion?: string, tem
           dispatch(getTemplate(id));
         }
         else {
-          dispatch(failureExistingTemplateUpdate(response.response));
+          dispatch(failureExistingTemplateUpdate());
         }
+      }).catch((error: any) => {
```
daniellemott: yay, no more spinners!

PR #142, opened 2020-03-12
Fix version bugs on Dashboard

daniellemott:
[AB#34057](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/08?workitem=34057)

This PR fixes multiple bugs related to template versions on the dashboard. 

- The dashboard was not displaying the most recent version of the template as its default
- Each adaptive card wrapper was not displaying the correct state (Published, Deprecated, or Draft) or the correct corresponding color) on the dashboard.


private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
```diff
@@ -24,6 +25,25 @@ interface Props {
   template: Template;
 }
 
+function retrieveStateValue(template: Template): string {
+  if (template.instances && template.instances[0] && template.instances[0].state) {
+    let state = template.instances[0].state;
+    switch (state) {
```
guptaharsh25: so a cleaner way to do this would be `return PostedTemplate.StateEnum[state]`, this returns a string for the enum value, so you can get rid of this switch statement

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
```diff
@@ -24,6 +25,25 @@ interface Props {
   template: Template;
 }
 
+function retrieveStateValue(template: Template): string {
```
guptaharsh25: I feel that we should put this function and other functions like this in a separate class inside the utils folder since we use something like this in several places.

also can we name the function something like `getLatestTemplateInstanceState` ?

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
```diff
@@ -48,13 +69,17 @@ class AdaptiveCardPanel extends React.Component<Props> {
           <TemplateStateWrapper style={{ justifyContent: "center" }}>
             <StatusIndicator
               state={
-                template.isLive
-                  ? PostedTemplate.StateEnum.Live
-                  : PostedTemplate.StateEnum.Draft
+                (template.instances && template.instances[0] && template.instances[0].state)
```
guptaharsh25: I feel that the case where `template.instances[0].state === undefined` should be handled in your `retrieveStateValue` function. this could be something like `<StatusIndicator state={retrieveStateValue(template)} ...`

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
guptaharsh25: to clarify, if `state = PostedTemplate.StateEnum.Draft`, `PostedTemplate.StateEnum[state] = "Draft"`

private-templates-service/client/src/components/Dashboard/PreviewModal/PreviewModal.tsx
```diff
@@ -47,11 +47,18 @@ interface State {
   templateVersion: string;
 }
 
+function getVersion(template?: Template): string {
```
guptaharsh25: put into utils class

guptaharsh25: you seem to already have this function in utils

private-templates-service/client/src/utils/getVersion/getVersion.ts
```diff
@@ -0,0 +1,8 @@
+import { Template } from 'adaptive-templating-service-typescript-node';
+
+export default function getVersion(template: Template): string {
```
guptaharsh25: rename to `getLatestVersion` maybe?

guptaharsh25: to add on to my previous comment about putting template utility function into utils, maybe we should make a file called `TemplateUtility.tsx` where we put all template related functions rather than having a file for each function.

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
daniellemott: Ooh thanks.

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
daniellemott: Done.

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
```diff
@@ -47,14 +50,16 @@ class AdaptiveCardPanel extends React.Component<Props> {
...
             <StatusIndicator
-              state={
-                template.isLive
-                  ? PostedTemplate.StateEnum.Live
-                  : PostedTemplate.StateEnum.Draft
+
+              state={template.instances && template.instances[0] && template.instances[0].state
+                ? template.instances[0].state
+                //  should never reach else statement
+                : PostedTemplate.StateEnum.Draft
```
kodyang: We are looking at the state of the `[0]`th element but we pass the `version` into `templateVersion`, is that correct?

See line 42 `<AdaptiveCard cardtemplate={template} templateVersion={version} />`

kodyang: I am thinking does `version` above always correspond to the item in the 0th index

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
```diff
@@ -47,14 +50,16 @@ class AdaptiveCardPanel extends React.Component<Props> {
...
               }
               style={{ marginRight: "10px" }}
             />
-            <Status>{template.isLive ? "Published" : "Draft"}</Status>
+            <Status>{state}
+            </Status>
```
kodyang: nit: keep on one line

private-templates-service/client/src/components/Common/VersionModal/VersionModal.tsx
```diff
@@ -115,27 +115,32 @@ class VersionModal extends React.Component<Props, State> {
...
+                    <VersionWrapper>
+                      {instance.version}
+                    </VersionWrapper>
+                    <DateWrapper>{instance.publishedAt ? getDateString(instance.publishedAt) : "Not published"}</DateWrapper>
+                    <StatusWrapper>
+                      <StatusIndicator state={
+                        instance.state
+                          ? instance.state
+                          // should never reach else statement
+                          : PostedTemplate.StateEnum.Draft
```
kodyang: If you know for sure that `instance.state` exists, then enter `instance.state!`

kodyang: It lets the type checker you are confident that instance.state is not null

guptaharsh25: that's cool, didn't know that

private-templates-service/client/src/components/AdaptiveCardPanel/AdaptiveCardPanel.tsx
daniellemott: From what I understood of what Grace told me, the template versions are stored in the `Template` object ordered from most recent (0th index) to least recent (last index). Since this is the Adaptive Card panel, we are always displaying the most recent version of the template. This is why the `state` is based off the template at the 0th index.

`version` is the actual string indicating the numeric value of the most recent template.

private-templates-service/client/src/components/Common/VersionModal/VersionModal.tsx
daniellemott: cool, thx!

PR #144, opened 2020-03-13
Fix bug preventing templating

daniellemott:
[AB#34058](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/09?workitem=34058)

Before this PR, users were unable to save a template with sample data due to a bug. This PR enables users to save sample data with their templates. Users will then be able to see templates bound with their data on the dashboard.


private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -193,6 +193,15 @@ export function updateTemplate(templateID?: string, currentVersion?: string, tem
     else {
       newTemplate.template = appState.currentTemplate.templateJSON;
     }
+
+    if (sampleDataJSON) {
+      // TODO: the clientSDK, backend, and DB adapter should be refactored such that the data is a json object and not an array of objects
```
kodyang: was the original plan to make the data an array of key-value pairs ?

private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -193,6 +193,15 @@ export function updateTemplate(templateID?: string, currentVersion?: string, tem
...
+      newTemplate.data = new Array(sampleDataJSON);
+    }
+    else {
+      newTemplate.template = appState.currentTemplate.sampleDataJSON;
```
kodyang: should this be 
`newTemplate.data = appState.currentTemplate.sampleDataJSON;`

private-templates-service/client/src/components/Common/AdaptiveCard/AdaptiveCard.tsx
```diff
@@ -47,29 +47,26 @@ export function renderAdaptiveCard(template: Template): any {
...
-function setContextRoot(data: string, context: ACData.EvaluationContext) {
+function setContextRoot(data: object, context: ACData.EvaluationContext) {
   try {
-    let dataString = JSON.stringify(data);
-    let dataJSON: JSON = JSON.parse(dataString);
-    context.$root = dataJSON;
+    context.$root = data;
   } catch (e) {
-    console.log(e);
   }
```
turalf: Since it is just an assignment now, I guess there is no need function as you are also not doing anything with try catch.

daniellemott: Removed!

private-templates-service/client/src/store/currentTemplate/actions.ts
daniellemott: I don't know. I think Grace + Seva might know. I think the data object should be changed to be a single JSON object, but I don't think that's in the scope of this PR.

private-templates-service/client/src/store/currentTemplate/actions.ts
daniellemott: My bad, that was a mistake. I've updated the logic. Could you please take another look?

PR #145, opened 2020-03-16
Multiple user bug fix

gracez72:
**Background**
The backend uses the access token passed from the client to identify which user is trying to access their templates/info. The server instantiates one instance of the TemplateServiceClient, which connects to the auth and storage provider, which serves multiple users.

The backend also needs to be installed directly by an user using it as a npm package and so needs to support passing in the access token once instead of every time a function is used. 

**Issue**
Previously user information was stored as a class variable in TemplateServiceClient which sometimes caused the wrong user's templates or info to be returned to the client because of callbacks in the backend.

**Changes**
This PR updates each public function in TemplateServiceClient to take an access token. This ensures that each response knows which user is accessing it, instead of relying on nodeJS' single threaded behaviour. 

_Additional Notes_
This PR also copied changes made by Seva's [PR](https://github.com/microsoft/adaptivecards-templates/pull/140), which removes unnecessary user info (first, last name, org, team) and can be merged in before or after.

**Updates to testing this PR locally**
Adding checks for the audience of the passed access token requires that we now run the backend locally to test PRs and update the client access token in `authproviders/AzureADProvider.ts` in the backend library.

### DIFF
--- a/private-templates-service/adaptivecards-templating-service/src/util/templateutils.ts
+++ b/private-templates-service/adaptivecards-templating-service/src/util/templateutils.ts
@@ -1,6 +1,4 @@
 import { ITemplate, ITemplateInstance, TemplateState } from "../models/models";
-import { version } from "mongoose";
-
 /**
  * @function
  * Updates passed ITemplate object to only have the latest version instance.
@@ -152,4 +150,13 @@ export function isValidJSONString(input: string) {
     return false;
   }
   return true;
+}
+
+/**
+ * Get token from authorization header.
+ * @param token 
+ */
+export function parseToken(token: string): string {
+  let bearer = token.split(/[ ]+/).pop();
+  return bearer || "";
 }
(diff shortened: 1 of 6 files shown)


private-templates-service/adaptivecards-templating-service/src/util/templateutils.ts
```diff
@@ -153,3 +151,12 @@ export function isValidJSONString(input: string) {
   }
   return true;
 }
+
+/**
+ * Get token from authorization header.
+ * @param token 
+ */
+export function parseToken(token: string): string {
+  let bearer = token.split(/[ ]+/).pop();
```
daniellemott: Should this be unit tested in case structure of token changes. Just a thought...

gracez72: Yep, I'll add some more unit tests for multiple users in a follow up PR and will include tests for malformed tokens/authorization headers!

PR #146, opened 2020-03-16
Configure cookies for HttpOnly and secure cookies

gracez72:
[TASK-33930](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/09?workitem=33930)

This PR changes server to require secure cookies and configures site cookies for HttpOnly for compliance requirements. It also enables strict HSTS headers to pass SSLScanner tests.

**Main file to review:** `server/src/app.ts`

### DIFF
--- a/private-templates-service/server/src/app.ts
+++ b/private-templates-service/server/src/app.ts
@@ -2,6 +2,7 @@ import express from "express";
 import path from "path";
 import passport from "./config/passport";
 import bodyParser from "body-parser";
+import session from "express-session";
 
 // import controllers
 import { TemplateServiceClient } from "../../adaptivecards-templating-service/src/TemplateServiceClient";
@@ -20,6 +21,13 @@ app.use(passport.initialize());
 app.use(passport.session());
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true }));
+app.use(session({
+  secret: '#{CLIENT_ID_TOKEN}#',
+  cookie: {
+    httpOnly: true,
+    secure: true
+  }
+}))
 
 let mongoDB = new MongoDBProvider({ connectionString: "#{DB_CONNECTION_TOKEN}#" });
 mongoDB.connect()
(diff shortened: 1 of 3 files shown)


private-templates-service/server/src/app.ts
```diff
@@ -20,6 +22,19 @@ app.use(passport.initialize());
 app.use(passport.session());
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true }));
+app.use(helmet.hsts({
+  maxAge: 15552000,
+  includeSubDomains: false
```
beervoley: Don't disable subDomains. We do not use any for now, but if somebody will want to host some part of the website on subdomain then hsts won't be enabled there :)

beervoley: Also, please remove **package.json** from the PR. 
run **git checkout origin/dev -- package.json** in the correct directory

private-templates-service/server/src/app.ts
```diff
@@ -20,6 +22,18 @@ app.use(passport.initialize());
 app.use(passport.session());
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true }));
+app.use(helmet.hsts({
+  maxAge: 15552000
```
kodyang: whats this

private-templates-service/server/src/app.ts
```diff
@@ -20,6 +22,18 @@ app.use(passport.initialize());
...
+}));
+app.use(session({
+  secret: '#{CLIENT_ID_TOKEN}#',
+  cookie: {
```
kodyang: is this a cookie we need compliance for

private-templates-service/server/src/app.ts
kodyang: oh 180 days

private-templates-service/server/src/app.ts
gracez72: Yeah this is for if we do use cookies to only use secure ones

PR #148, opened 2020-03-17
Edit a template name from the preview page

kodyang:
## Description
[AB#33852](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/33852) This lets you change a template name from the preview page.

## Screenshots
I didn't have any designs to work off of but It is consistent with the rest of the site.
### From preview page:
image

### Modal:
image


private-templates-service/client/src/components/Common/EditNameModal/EditNameModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (name) {
+      props.updateName(name);
+    }
+    props.closeModal();
+  }
+
+  return (
+    <BackDrop>
+      <Modal>
+        <Header>Edit Card Name</Header>
```
markAtMicrosoft: Literal strings should be in your string resource file.

private-templates-service/client/src/components/Common/EditNameModal/EditNameModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+const EditNameModal = (props: Props) => {
+  const [name, setName] = useState(props.currentName);
+
+  const onChange = (event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
+    setName(newValue || '');
+  }
+
+  const updateName = () => {
+    if (name) {
+      props.updateName(name);
```
guptaharsh25: can we add a check to see if the name changed or not before calling `props.updateName(name)`?

private-templates-service/client/src/components/Common/EditNameModal/EditNameModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+    props.closeModal();
+  }
+
+  return (
+    <BackDrop>
+      <Modal>
+        <Header>Edit Card Name</Header>
+        <Description>The card will still be accessible at the same URLs.</Description>
+        <EditWrapper onSubmit={updateName} >
```
guptaharsh25: not sure I understand what `onSubmit` does here. I think it just wraps the `StyledInput` but isn't calling `updateName` when `PrimaryButton` is clicked enough?

private-templates-service/client/src/components/Common/EditNameModal/styled.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+
+export const EditWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  padding-bottom: 24px;
+`;
+
+export const StyledInput = styled(TextField)``;
```
guptaharsh25: Do we need `StyledInput` if we're not adding any CSS to it? Can we just use `Textfield`?

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -71,7 +90,7 @@ const NavBar = (props: NavBarProps) => {
         <Banner>
           <MobileBanner>
             <StyledLogo src={Logo} />
-            <Styledh1>{props.template ? props.template.name : props.currentPageTitle}</Styledh1>
+            <Styledh1>{props.templateName ? props.templateName : props.currentPageTitle}</Styledh1>
```
guptaharsh25: I think we should try to take out references to `templateName` and `templateJSON` from the redux store and access them from the `template` object in the store so that there's one true source of template info. What do you think?

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -96,14 +115,15 @@ const NavBar = (props: NavBarProps) => {
     default:
       return (
         <Banner>
-          <MobileBanner>
+          <BaselineBanner>
             <StyledLogo src={Logo} />
-            <Styledh1>{props.currentPageTitle}</Styledh1>
-          </MobileBanner>
+            <Styledh1>{props.templateName || props.currentPageTitle}</Styledh1>
+            {!props.isFetching && <EditButton onClick={editName} iconProps={{ iconName: 'Edit' }} />}
```
guptaharsh25: here, are you saying that the default case will be used for the template preview page? if so, does this imply that name is only editable from that page? do we not want to allow name change from designer page?

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -96,14 +115,15 @@ const NavBar = (props: NavBarProps) => {
     default:
       return (
         <Banner>
-          <MobileBanner>
+          <BaselineBanner>
```
guptaharsh25: why do we need `BaselineBanner`? Seems like the only difference is `align-items: baseline;`

private-templates-service/client/src/store/currentTemplate/reducers.ts
```diff
@@ -37,7 +37,7 @@ export function currentTemplateReducer(state = initialState, action: CurrentTemp
         ...state,
         templateID: "",
         templateJSON: require('../../assets/default-adaptivecards/defaultAdaptiveCard.json'),
-        templateName: "Untitled",
+        templateName: "Untitled Card",
```
guptaharsh25: Should this be moved to the strings file? idk

private-templates-service/client/src/components/Common/EditNameModal/styled.tsx
Jackbk: whats the ``?

private-templates-service/client/src/components/Common/EditNameModal/EditNameModal.tsx
kodyang: great idea

private-templates-service/client/src/components/Common/EditNameModal/EditNameModal.tsx
kodyang: since editwrapper is a `form` then the `submit` event is triggered when the enter button is pressed when the focus is on a child component of the form.

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: done

private-templates-service/client/src/components/Common/EditNameModal/styled.tsx
kodyang: removed

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: khang is working on that part

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: the pencil icon is a few pixels off without that style lol

private-templates-service/client/src/store/currentTemplate/reducers.ts
kodyang: yeah

private-templates-service/client/src/components/NavBar/NavBar.tsx
guptaharsh25: okay sure. but i still feel like we should add a new case in this switch statement to handle the PreviewModal page instead of using the default case.

private-templates-service/client/src/utils/requireShared/requireShared.tsx
```diff
@@ -57,7 +57,7 @@ const requireShared = <P extends object>(Component: React.ComponentType<P>) => {
     render() {
       return (
         <React.Fragment>
-          {this.isShared() ?
+          {!this.isShared() ?
```
guptaharsh25: Are you sure about this?

private-templates-service/client/src/components/NavBar/NavBar.tsx
kodyang: agreed, done

private-templates-service/client/src/utils/requireShared/requireShared.tsx
kodyang: nice catch, i was using this for testing.

PR #156, opened 2020-03-17
Add automatic logout after 30 minutes of inactivity

beervoley:
This pr addresses [ACB34148](git push --set-upstream origin users/vsevolod/34148-log-out) about logging the user out after 30 minutes of idle time.
It also fixes storage providers tests and InMemoryStorage providers based on models update in https://github.com/microsoft/adaptivecards-templates/pull/147
To test it locally with idle time less than 30 minutes:
Update `timeout` in `client/src/App.tsx`
image

### DIFF
--- a/private-templates-service/client/src/App.tsx
+++ b/private-templates-service/client/src/App.tsx
@@ -25,6 +25,9 @@ import config from "./Config";
 import "bootstrap/dist/css/bootstrap.css";
 import { OuterAppWrapper, MainAppWrapper, MainApp } from "./styled";
 
+// Constants
+import Constants from "./globalConstants"
+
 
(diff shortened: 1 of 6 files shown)


private-templates-service/client/src/App.tsx
```diff
@@ -110,6 +113,10 @@ class App extends Component<Props, State> {
 
     return (
       <Router>
+        <IdleTimer
+          element={document}
+          onIdle={this.onIdle}
+          timeout={1800000} />
```
turalf: keep it in a separate constants file/collection?

beervoley: Done :)

PR #158, opened 2020-03-18
Log when user data is accessed

beervoley:
This PR addressed part of the [ACB33922](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/33922) compliance ticket related to logging. Now we'll log every time personal user data (oid) is accessed.

### DIFF
--- a/private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
+++ b/private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
@@ -196,7 +196,6 @@ export class TemplateServiceClient {
    * @param token 
    */
   public async getUser(token?: string): Promise<JSONResponse<IUser>> {
-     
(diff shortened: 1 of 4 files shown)


private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -186,6 +185,8 @@ export class TemplateServiceClient {
     if (!result.success){
       return { success: false, errorMessage: ServiceErrorMessage.UserNotFound }
     }
+    logger.info(`User with oid ${authId} requested data.`);
```
gracez72: Do we need to log what data was requested and returned?

beervoley: Looks like we don't:
image

PR #161, opened 2020-03-18
Enabling sending e-mails when sharing template

guptaharsh25:
Quick PR for enabling sending emails when sharing.

Also did some refinement for the Sharing Modal to ensure it matches the designs.

image


smmatte:
To confirm, what happens if a person doesn't have Outlook? what happens on Mac, Linux? Should the behavior and text not be about Open Default Mail Client?

private-templates-service/client/src/components/Common/ShareModal/ShareModalForm/ShareModalForm.tsx
```diff
@@ -31,23 +40,9 @@ const mapDispatchToProps = (dispatch: any) => {
...
-    this.handleChange = this.handleChange.bind(this);
-    this.handleSubmit = this.handleSubmit.bind(this);
-  }
-
-  handleChange(event: any) {
-    this.setState({ value: event.target.value });
-  }
+class ShareModalForm extends React.Component<ShareModalFormProps> {
 
   handleSubmit(event: any) {
```
kodyang: how does this work without `this.handleSubmit.bind(this);`

can you change it into `handleSubmit = () => { ... }`

Also, since you are no longer using a form there is no need for `event.preventDefault()` so you can get rid of `event` in the params

private-templates-service/client/src/components/Common/ShareModal/styled.tsx
```diff
@@ -93,3 +93,11 @@ export const ButtonGroup = styled.div`
 export const CancelButton = styled(DefaultButton)`
   margin-right: 12px;
 `;
+
+export const SendMailButton = styled(DefaultButton)`
+  max-width: fit-content;
```
kodyang: please add this to all buttons in this file:
```
  &: focus {
    outline: 0px;
  }
```

gets rid of the ugly square when u click on a button.

private-templates-service/client/src/components/Common/ShareModal/ShareModalForm/ShareModalForm.tsx
guptaharsh25: addressed

private-templates-service/client/src/components/Common/ShareModal/styled.tsx
guptaharsh25: addressed

PR #163, opened 2020-03-19
fixed styling for footer

guptaharsh25:
[AB#34162](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34162)

just a quick styling fix for the footer

### DIFF
--- a/private-templates-service/client/src/components/Dashboard/styled.tsx
+++ b/private-templates-service/client/src/components/Dashboard/styled.tsx
@@ -16,6 +16,7 @@ export const Title = styled.h1`
 
 export const OuterDashboardContainer = styled.div`
   display: contents;
+  height: calc(100vh - 61px);
 `;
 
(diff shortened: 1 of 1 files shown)


private-templates-service/client/src/components/Dashboard/styled.tsx
```diff
@@ -31,7 +32,8 @@ export const OuterWindow = styled.div`
   flex-direction: row;
   align-items: stretch;
   justify-content: space-between;
-  height: 100%;
+  height: auto;
```
daniellemott: do we need height: auto?

guptaharsh25: yeah we do, the dashboard contents were overflowing out of this container when the browser was resized to be smaller, which was causing the bug

PR #164, opened 2020-03-19
Check token expiration date in backend

gracez72:
[TASK-34160](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/09?workitem=34160)

This PR updates the access token passed to the backend to be the one that is rotated automatically if it has expired. The expiration date of the token is now also checked in the backend and will return 401 if the token has expired.


private-templates-service/client/src/store/currentTemplate/actions.ts
```diff
@@ -319,7 +319,7 @@ function initClientSDK(dispatch: any, getState: () => RootState, ): TemplateApi
   const api = new TemplateApi();
   const state = getState();
   if (state.auth.accessToken) {
-    api.setApiKey(0, `Bearer ${state.auth.accessToken!.idToken.rawIdToken}`);
+    api.setApiKey(0, `Bearer ${state.auth.accessToken!.accessToken}`);
```
guptaharsh25: on line 279 there is a call to make `const api = initClientSDK(dispatch, getState);` which is a helper method at the bottom of the file. can you make the setApiKey changes in there so that its consistent for all client SDK related actions?

private-templates-service/adaptivecards-templating-service/src/authproviders/AzureADProvider.ts
```diff
@@ -44,8 +44,11 @@ export class AzureADProvider implements AuthenticationProvider {
       }
     }
 
+    // Check expiry date on token
+    result = new Date() <= new Date(decodedToken.payload.exp * 1000); 
+
     // Check aud of token matches the client ID of env app
-    result = "#{CLIENT_ID_TOKEN}#" === decodedToken.payload.aud;
+    result = "#{CLIENT_ID_TOKEN}#" === decodedToken.payload.appid || "#{CLIENT_ID_TOKEN}#" === decodedToken.payload.aud;
```
guptaharsh25: are we going to have to change both of these now to make the app run locally?

private-templates-service/client/src/store/currentTemplate/actions.ts
guptaharsh25: so after going through all the changes in this PR, I feel like we can make this change in that helper method `initClientSDK`, but we can also put it in a util file and call that method in all the actions that require changes in this PR.

gracez72: Done, thanks!

private-templates-service/adaptivecards-templating-service/src/authproviders/AzureADProvider.ts
gracez72: Moved to client id token to variable so it only needs to be changed once

private-templates-service/adaptivecards-templating-service/src/authproviders/AzureADProvider.ts
```diff
@@ -44,8 +44,12 @@ export class AzureADProvider implements AuthenticationProvider {
       }
     }
 
+    // Check expiry date on token
+    result = new Date() <= new Date(decodedToken.payload.exp * 1000); 
+
     // Check aud of token matches the client ID of env app
-    result = "#{CLIENT_ID_TOKEN}#" === decodedToken.payload.aud;
+    let appId = "#{CLIENT_ID_TOKEN}#";
+    result = appId === decodedToken.payload.appid || appId === decodedToken.payload.aud;
```
beervoley: After expiry date was checked on the token, the result of that check is not used. Modify the logic so if expiry date check failed, function will return false.

PR #167, opened 2020-03-20
Add owner name and picture

daniellemott:
[#AB34149](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/10?workitem=34149) Display the Correct Owner Name and Picture on Dashboard

What this PR does:
- Uses a user's oID to request the corresponding owner's picture and name from Microsoft Graph.
- If the user doesn't have a profile picture, it displays the first letter of their first name in a circle.
- Displays the user's profile picture and name in the Recently Viewed table on the dashboard.

**Note**
- The `Error loading templates` message will disappear as soon as we clear the database. 
- The owner's picture and name must be added in the template info page as well. I decided to separate them into two PRs. 

image

If the last edited user isn't the person who's logged in:
image


private-templates-service/client/src/store/templateOwner/reducers.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function templateOwnerReducer(state = initialState, action: GetOwnerNameAction | GetOwnerProfilePictureAction): OwnerState {
+  switch (action.type) {
+    case GET_OWNER_NAME:
+    case GET_OWNER_PROFILE_PICTURE:
+      return {
+        ...state,
+        isFetching: true,
+      };
```
beervoley: for case **GET_OWNER_NAME** if there's no code, it will just use the logic of the next case (GET_OWNER_PROFILE_PICTURE)?

private-templates-service/client/src/store/templateOwner/reducers.ts
```diff
@@ -0,0 +1,64 @@
...
+    case GET_OWNER_NAME_FAILURE:
+    case GET_OWNER_PROFILE_PICTURE_FAILURE:
+      return {
+        ...state,
+        isFetching: false,
+      }
+    case GET_OWNER_NAME_SUCCESS:
+    case GET_OWNER_PROFILE_PICTURE_SUCCESS:
```
beervoley: same as above

private-templates-service/client/src/store/templateOwner/reducers.ts
daniellemott: yes, that's right! the idea is that GET_OWNER_NAME is returning {...state, isFetching: true,} as well

beervoley: alright:) I know it's repetition of code, but wouldn't it be easier to read if you just put case logic there anyway?:)

private-templates-service/client/src/store/templateOwner/reducers.ts
```diff
@@ -0,0 +1,64 @@
...
+const initialState: OwnerState = {
+  owner: undefined,
+  isFetching: false,
+}
+
+const initialOwnerState: OwnerType = {
+  displayName: "",
+}
+
+function owner(owner = initialOwnerState, action: GetOwnerNameAction | GetOwnerProfilePictureAction): OwnerType | undefined {
```
beervoley: Can we rename this function to indicate what it's doing?:)

daniellemott: sure!

private-templates-service/client/src/store/templateOwner/reducers.ts
daniellemott: how about adding a comment instead? :)

beervoley: sure:)

private-templates-service/client/src/components/Common/ShareModal/ShareModal.tsx
```diff
@@ -10,7 +10,7 @@ import ModalHOC from '../../../utils/ModalHOC';
 import { closeModal } from '../../../store/page/actions';
 import { connect } from 'react-redux';
 
-import { TextField, Button } from 'office-ui-fabric-react';
+import { TextField } from 'office-ui-fabric-react';
```
kodyang: thanks for removing unused imports :)

private-templates-service/client/src/store/templateOwner/actions.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const client = getAuthenticatedClient(state.auth.accessToken);
+    return client.api('/users/' + oID + '/photo/$value').get()
+      .then((image: Blob) => {
+        const imageURL = URL.createObjectURL(image);
+        dispatch(requestOwnerProfilePictureSuccess(imageURL));
+      },
+        (fail: any) => {
+          dispatch(requestOwnerProfilePictureFailure());
+        })
```
kodyang: indentation

private-templates-service/client/src/store/templateOwner/reducers.ts
```diff
@@ -0,0 +1,65 @@
...
+  switch (action.type) {
+    case GET_OWNER_NAME_SUCCESS:
+      return {
+        ...owner,
+        displayName: action.ownerName
+      }
+    case GET_OWNER_PROFILE_PICTURE_SUCCESS:
+      return {
+        ...owner,
+        imageURL: action.ownerImageURL,
```
kodyang: Question about this.

It seems like each owner profile picture component is updating this single value. If there are multiple templates from different people does it show each person's different picture or just the last one fetched from the server? I would have expected the URLs to be stored in some sort of array or dict.

daniellemott: If you refer to line 30 in RecentlyViewedTable.tsx, you will see the LastEditedUserInfo component (formerly known as OwnerInfo) is constructed for each Template in the RecentlyViewed array. This is all based on the oID in each Template in the RecentlyViewed. The person's profile picture should update accordingly.

private-templates-service/client/src/store/templateOwner/actions.ts
daniellemott: my linter auto corrects to this indentation?

kodyang: right, sorry I should have been more clear. can the callback be put on the same line like in line 69? 

`}, (fail: any) => {`

private-templates-service/client/src/store/templateOwner/reducers.ts
kodyang: Right. I believe each user's image is fetched but the URL is stored at this location in the redux store. Each component looks at this element in the store to show the image. I still think the images may not be displayed correctly, but I think the best way to verify is to try it out. 

I tried to run it locally with two different accounts but I'm not able to see published templates from other users, preventing me from verifying this.

daniellemott: Fixed it! 
image

private-templates-service/client/src/store/templateOwner/reducers.ts
```diff
@@ -0,0 +1,72 @@
...
+    default:
+      return owner;
+  }
+}
+
+export function templateOwnerReducer(state = initialState, action: GetOwnerNameAction | GetOwnerProfilePictureAction): OwnerState {
+  switch (action.type) {
+    // empty case statement will "fall through" to logic in the next non-empty case
+    case GET_OWNER_NAME:
+    case GET_OWNER_PROFILE_PICTURE:
```
kodyang: I think on `GET_OWNER_NAME` and `GET_OWNER_PROFILE_PICTURE`, it makes sense to clear `displayNames` and `imageURLs`

kodyang: like set them to `{}`

daniellemott: Oh right, thanks for the reminder. Done!

private-templates-service/client/src/store/templateOwner/actions.ts
```diff
@@ -52,6 +54,13 @@ function requestOwnerProfilePictureFailure(): GetOwnerProfilePictureAction {
   }
 }
 
+export function ClearOwners(): ClearOwnersAction {
+  console.log("clear");
```
kodyang: remove console log

daniellemott:
> i prefer clearing on GET because then the method can be used outside the context of the dashboard, but its perfectly fine as is.

Okay, I may change it in the next PR but going to leave as is right now.

PR #174, opened 2020-03-24
Fix adding tag creates new template version bug

gracez72:
**Ticket**: [TASK-34210](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/10?workitem=34210)

**Bug Description**
Adding/removing a new tag causes a new draft version to be added when the template is live or deprecated. 

**Fix**
This is partly because the client sends the template json in the request. This PR removes the template json if the template is only being updated and adds a check in the backend to prevent new versions from being added when tags or data are added to a template. 

This PR also reverts the pink delete button change because the delete modal was added.


private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -401,11 +403,14 @@ export class TemplateServiceClient {
...
               templateInstances.push(instance);
-              // Pushing existing deprecated version back
-              version = incrementVersion(existingTemplate);
-              templateInstance = setTemplateInstanceParam(templateInstance, templateData, templateState, isShareable, version);
-              templateInstances.push(templateInstance);
+              if (template || state) {
+                version = incrementVersion(existingTemplate);
+                templateInstance = setTemplateInstanceParam(templateInstance, templateData, templateState, isShareable, version);
+                templateInstances.push(templateInstance);
+              }
```
beervoley: in this **else** block code is repeated from the first **if** block. May be try to refactor it? Either put it both under one if statement with **or** condition or extract it to a method?

gracez72: Done, thanks!

PR #176, opened 2020-03-25
Address accessibility issues

beervoley:
Addressed:
[BUG34218](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34218)
[BUG34220](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34220)
[BUG34221](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34221)
[BUG34223](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34223)
[TASK34236](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34236)

### DIFF
--- a/private-templates-service/client/src/components/AdaptiveCardPanel/styled.tsx
+++ b/private-templates-service/client/src/components/AdaptiveCardPanel/styled.tsx
@@ -61,7 +61,7 @@ export const TemplateName = styled.div`
 export const TemplateUpdatedAt = styled.div`
   font-family: Segoe UI Regular;
   font-size: 0.8rem;
-  color: grey;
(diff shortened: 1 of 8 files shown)


beervoley:
> is there a reason you added the `aria-label` in the styled.tsx?
> 
> I think its best for all non-styling to happen in the jsx files. can you move them to the tags themselves?

Okay, moved to the tsx files itself. I just thought that it's part of the styling and hiding it in styled.tsx would be nicer and cleaner.

PR #179, opened 2020-03-26
Server Side Binding

daniellemott:
[AB#33991](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/10?workitem=33991) Server Side Binding (Developer Feature)

What this PR does:
- Enables "server side binding"
- If a developer makes a successful POST request to the `template/{id}` endpoint with a data JSON  and `"bindData": "true"`in the body of the request , it will return the template corresponding to the id with the data that was passed in bound to the template 
- If the version is passed in the body, it will return the corresponding template version with the data JSON bound. If the version is not passed in, it will return the latest version of that template with the data JSON bound

I ran the server locally and used Postman to test. 

_Clarification of the term "data binding"_:

**If** the template JSON stored in the DB is: 
`{
    text: "{name}"
}`
**and** the data JSON passed in is: 
`{
   "name": "danielle"
}`

**and** a correct GET request is made following the steps listed above, **then** the template JSON field returned will be:
`{
   text: "danielle"
}`


private-templates-service/adaptivecards-templating-service/src/util/templateutils.ts
```diff
@@ -159,4 +161,18 @@ export function isValidJSONString(input: string) {
...
+export function createCard(template: JSON, data: JSON): JSON {
+  try {
+    let dataTemplate: ACData.Template = new ACData.Template(template);
+    let context: ACData.EvaluationContext = new ACData.EvaluationContext();
+    context.$root = data;
+    let card: JSON = dataTemplate.expand(context);
+    return card;
+  }
+  catch {
+    return JSON.parse('{}');
```
turalf: Would it be different than "return {}";

private-templates-service/adaptivecards-templating-service/package.json
```diff
@@ -16,12 +16,15 @@
...
   "dependencies": {
+    "adaptivecards-templating": "^0.1.1-alpha.1",
     "axios": "^0.19.2",
     "bunyan": "^1.8.12",
     "express": "^4.17.1",
     "express-validator": "^6.4.0",
+    "install": "^0.13.0",
     "jws": "^4.0.0",
     "mongoose": "^5.8.10",
+    "npm": "^6.14.4",
```
kodyang: I am not sure you intended to include these ?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -717,7 +719,22 @@ export class TemplateServiceClient {
...
+      if (!version) {
+        if (data) {
+          let resultTemplates: ITemplate[] = [];
+          let template: ITemplate = templates[0];
+          if (!template.instances || !template.instances[0]) {
+            return { success: false, errorMessage: "No template instances found" };
+          }
+          let boundJSON: JSON = createCard(template.instances[0].json, data);
+          template.instances![0].json = boundJSON;
+          //return only the most recent version of the template with data bound
```
kodyang: If hit the endpoint specifying a version, will it bind that version or the most recent version

private-templates-service/adaptivecards-templating-service/src/tests/router.test.ts
```diff
@@ -866,4 +866,66 @@ describe("Basic Get Templates", () => {
     // Check that only the latest version is returned
     expect(template.instances[0].version).toMatch("1.2");
   });
+
+  // Authenticated post request
+  it("pass in a data json and id and should receive a template with the data bound", async () => {
```
kodyang: Is it possible to add tests to bind data to a template with multiple versions? I think this test file should have a testUtils so you can call a helper function to create these templates for you.

private-templates-service/adaptivecards-templating-service/src/util/templateutils.ts
daniellemott: I tried to do that, but the type checker wouldn't let. me :(

private-templates-service/adaptivecards-templating-service/package.json
daniellemott: I did not, thanks. Removed them.

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: It will bind that version, not the most recent version.

private-templates-service/adaptivecards-templating-service/src/tests/router.test.ts
daniellemott: There's no testUtils page, but I added to the unit tests.

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1174,6 +1262,32 @@ export class TemplateServiceClient {
...
+          token,
+          req.params.id,
+          req.body.data,
+          req.body.version,
+          isClient
+        )
+
+        if (!response.success || (response.result && response.result.length === 0)) {
+          const err = new TemplateError(ApiError.DataBindingFailed, "Data binding failed.");
+          return res.status(404).json({ error: err });
```
daniellemott: I wasn't sure about which error codes to use?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1174,6 +1262,32 @@ export class TemplateServiceClient {
...
+        }
+        return res.status(200).json({ templates: response.result });
```
daniellemott: Is 200 ok?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
gracez72: Can return 400 bad request

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
gracez72: 200 is ok!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,92 @@ export class TemplateServiceClient {
...
+   * If a version is passed in, it will update the template JSON of that version.
+   * If not version is passed, it will update the latest version.
+   * @param {string} templateId - unique template id
+   * @param {JSON} data - data json to bound with template json
+   * @param {string} version - version number, used with templateId
+   * @param {boolean} isClient - used to ignore updating the hit number on a template
+   * @returns Promise as valid json
+   */
+  public async bindData(
+    token: string,
```
gracez72: for public functions token should be optional so that we can use the default one stored in the authprovider class

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,92 @@ export class TemplateServiceClient {
...
+        //return only the most recent version of the template with data bound 
+        template.instances = [template.instances[0]];
+        resultTemplates.push(template);
+        return { success: true, result: resultTemplates };
+      }
+    }
+
+
+    // Filter for the latest template version (instance)
+    let resultTemplates: ITemplate[] = [];
```
gracez72: Do we need this block? Searching for a specific version can probably be done in the same block as getting the latest version. We can sort the versions and if version is not defined, directly push the first element (latest version) to the resulting instances array, otherwise continue searching for the given version

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,92 @@ export class TemplateServiceClient {
...
+      _id: templateId,
+    };
+
+    let response = await this.storageProvider.getTemplates(templateQuery, undefined, undefined);
+
+    if (!response.success || !response.result) return response;
+
+    let templates: ITemplate[] = response.result;
+
+    if (templateId && templates.length > 0) {
```
gracez72: templateid must be defined based on the function params, we can probably add a check that the templates.length === 0 before we proceed (since only one template object should be returned from the database because we pass in the id)

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,92 @@ export class TemplateServiceClient {
...
+      if (templates![0].owner !== userId && templates![0].isLive === false) return { success: true, result: [] };
```
gracez72: Can remove this check since we'll be getting rid of isLive for the template object and any user can access the template even if they're not the owner

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,92 @@ export class TemplateServiceClient {
...
+        template.instances = [template.instances[0]];
+        resultTemplates.push(template);
+        return { success: true, result: resultTemplates };
+      }
+    }
+
+
+    // Filter for the latest template version (instance)
+    let resultTemplates: ITemplate[] = [];
+    for (let template of templates) {
```
gracez72: We can assume the resulting templates has only one object, we should fail if it doesn't have exactly one element since we searched by templateid

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1174,6 +1262,32 @@ export class TemplateServiceClient {
       let tags: string[] | string = req.body.tags;
       let data: JSON[] | JSON = req.body.data;
 
+      let bindData: boolean | undefined = req.body.bindData ? req.body.bindData.toLowerCase() === "true" : undefined;
```
gracez72: Are we adding this functionality to the client sdk? If yes do you mind also adding it to the swagger definition file?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -1174,6 +1262,32 @@ export class TemplateServiceClient {
...
+
+      if (bindData) {
+        if (req.body.data !== undefined && (!(req.body.data instanceof Object) || !isValidJSONString(JSON.stringify(req.body.data)))) {
+          const err = new TemplateError(ApiError.InvalidTemplate, `Data must be valid JSON.`);
+          return res.status(400).json({ error: err });
+        }
+        let isClient: boolean | undefined = req.query.isClient ? req.query.isClient.toLowerCase() === "true" : undefined;
+        let token = parseToken(req.headers.authorization!);
```
gracez72: token is defined above as well, can remove!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: kk!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: Nope, will remove. I didn't think about the fact that it's only returning one template!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: Do you mean check that templates.length === 1?

gracez72: Oh yep my bad check templates.length === 1!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: As discussed, let's leave it out for right now. I don't see the need.

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,83 @@ export class TemplateServiceClient {
...
+      _id: templateId,
+    };
+
+    let response = await this.storageProvider.getTemplates(templateQuery, undefined, undefined);
+
+    if (!response.success || !response.result) return response;
+
+    let templates: ITemplate[] = response.result;
+
+    if (templates.length != 1) {
```
gracez72: use !==?

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -791,6 +793,83 @@ export class TemplateServiceClient {
...
+    }
+
+    sortTemplateByVersion(templates![0]);
+
+    let resultTemplates: ITemplate[] = [];
+    let template = templates[0];
+    if (template.isLive === false && template.owner !== userId) return { success: false, errorMessage: "Invalid user ID" };
+    if (!template.instances) { return { success: false, errorMessage: "Invalid template ID" } };
+    if (version) {
+      for (let instance of template.instances) {
```
gracez72: There's a bit of repetition with the creating card, editing json and pushing the template instance - can maybe edit to something like this? 
```
  // Default latest version
    let selectedInstance = template.instances![0];
    if (!version) {
      for (let instance of template.instances) {
        if (instance.version === version) {
          selectedInstance = instance;
          break;
        }
      }
    }
    let boundJSON: JSON = createCard(selectedInstance.json, data);
    selectedInstance.json = boundJSON;
    template.instances = [selectedInstance];
    return { success: true, result: [template] }  
```

PR #183, opened 2020-03-26
Tags Spinner

kodyang:
## Description
Many users have commented that adding a tag causes the whole page to turn into a spinner, they said it would be best to only have the tags section be a spinner. 

[AB#34265](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34265)

## Video
Sorry, my internet is slow. It taking a long time has nothing to do with this change!
Adaptive-Cards-CMS-and-8-more-pa


kodyang:
Hey Rachel, we update the backend with the new list of tags and wait for the response before showing the tags again. In this time we show a spinner. In theory, something could go wrong and the tags aren't actually updated

PR #185, opened 2020-03-26
 fixing save bug

nnmkhang:
[AB#34271](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34271)

Please Re-Review. 

Overview: 
- Added SpinnerModal
- Added URLs for the designer 
- Changed ComponentDidMount logic in designer.
- Spinner now overlays when an action is in progress in the designer (ie Save) 

Snapshots ( Pls notice the urls ) :

Saved Card Before:
image

Saved Card After: 
image

New Card before:
image

New Card After:
image

Spinner:
image

Spinner after removing backgroud: 
image


private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -149,7 +149,7 @@ function initDesigner(): ACDesigner.CardDesigner {
 }
 
 function onSave(designer: ACDesigner.CardDesigner, props: DesignerProps): void {
-  if(props.templateID === ""){
+  if(props.templateID === "" || props.templateID === undefined){
```
beervoley: Can substitute boolean expression to `if (not props.templateID)`

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -100,8 +100,8 @@ const NavBar = (props: NavBarProps) => {
         <Banner>
           <MobileBanner>
             <StyledLogo src={Logo} />
-            <Styledh1>{(props.templateID === "" && UNTITLEDCARD) || props.templateName}</Styledh1>
-            {props.templateID !== "" && <EditButton onClick={editName} iconProps={{ iconName: 'Edit' }} />}
+            <Styledh1>{((props.templateID === "" || props.templateID === undefined) && UNTITLEDCARD) || props.templateName}</Styledh1>
+            {(props.templateID !== "" && props.templateID !== undefined ) && <EditButton onClick={editName} iconProps={{ iconName: 'Edit' }} />}
```
beervoley: Can also substitute boolean expression same as comment above

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+    super(props);
+  }
+  
+  componentDidUpdate(){
+    if(!this.props.isFetching){
+      if(this.props.closeAction){
+        this.props.closeAction();
+      }
+      this.props.closeModal();
+    }
```
kodyang: Why does this spinner modal do actions passed in by parent components?

I think it makes sense for a spinner to be UI only, the parent component should handle side effects.

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -60,18 +66,43 @@ interface State {
...
+  
+  updateURL = () => {
+      this.props.history.push("/designer/"+ this.props.templateID + "/" + this.props.version);
+  }
+  
+  toggleSpinner = (isFetching: boolean) => {
+    if(isFetching){
+      this.props.openModal(ModalState.Spinner)
+    }
+  }
```
kodyang: Why do we do this:
`isFetching === true -> component did update -> dispatch action to open spinner modal -> ModalState === spinner -> spinner open`

Why cant we do this
`isFetching === true -> spinner open`
By extension, `isFetching === false -> spinner not open`

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -122,9 +153,10 @@ class Designer extends React.Component<DesignerProps,State> {
     return (
       <React.Fragment>
         <DesignerWrapper id="designer-container" />
-        {this.props.modalState===ModalState.Save && <SaveModal designerSampleData = {designer.sampleData} designerTemplateJSON = {designer.getCard()}/>}
-        {this.props.modalState === ModalState.EditName && <EditNameModal />}
-      </React.Fragment>
+        {this.props.modalState === ModalState.Spinner && <SpinnerModal closeAction = {this.updateURL}/>}
```
kodyang: could just make this 
`isFetching && <spinnermodal ...`

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -101,9 +101,9 @@ const NavBar = (props: NavBarProps) => {
       return (
         <Banner>
           <MobileBanner>
-            <StyledLogo aria-label={STRINGS.LOGO_DESCRIPTION} src={Logo} />
-            <Styledh1>{(props.templateID === "" && STRINGS.UNTITLEDCARD) || props.templateName}</Styledh1>
-            {props.templateID !== "" && <EditButton onClick={editName} iconProps={{ iconName: 'Edit' }} />}
+            <StyledLogo src={Logo} />
```
kodyang: why did we get rid of aria label

kodyang:
The spinner is hard to see, maybe have a small white square behind it

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
guptaharsh25: This was my idea. Do you have ideas for how to handle this logic in the parent component?

@nnmkhang maybe we do something like `prevProps.isFetching === true && this.props.isFetching === false` inside of `componentDidUpdate()` ? I think that might work, you can try it

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+  constructor(props: Props) {
+    super(props);
+  }
+  
+  componentDidUpdate(){
+    if(!this.props.isFetching){
+      if(this.props.closeAction){
+        this.props.closeAction();
+      }
+      this.props.closeModal();
```
guptaharsh25: I agree with what Kody said in his comments, although we treat this component like a Modal it really isn't, it's just an overlay for when `isFetching === true`. I feel like the `openModal()` and `closeModal()` actions aren't needed in this case.

guptaharsh25: Also that implies we don't need the `ModalState` enum

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -15,6 +15,8 @@ import { DesignerWrapper } from './styled';
 import EditNameModal from '../Common/EditNameModal';
 import SaveModal from './SaveModal/SaveModal';
 import { ModalState } from '../../store/page/types';
+import { withRouter, RouteComponentProps } from 'react-router-dom';
+import SpinnerModal from '../Common/SpinnerModal';
```
guptaharsh25: Could you group similar imports together?

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -60,18 +66,43 @@ interface State {
...
+  }
+  
+  toggleSpinner = (isFetching: boolean) => {
+    if(isFetching){
+      this.props.openModal(ModalState.Spinner)
+    }
+  }
+
+  componentDidUpdate() {
+    this.toggleSpinner(this.props.isFetching);
```
guptaharsh25: won't this get triggered if ANY prop changes? i could be interpreting it wrong

private-templates-service/client/src/components/Designer/SaveModal/SaveModal.tsx
```diff
@@ -87,7 +87,7 @@ class SaveModal extends React.Component<Props, State> {
       this.setState({ templateName: newValue });
     }
   }
-
+
```
guptaharsh25: remove unnecessary whitespace?

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -101,9 +101,9 @@ const NavBar = (props: NavBarProps) => {
...
+            <Styledh1>{((props.templateID === "" || props.templateID === undefined) && STRINGS.UNTITLEDCARD) || props.templateName}</Styledh1>
```
guptaharsh25: Correct me if I'm wrong but when the REST call is in progress I believe the template name becomes blank, can we handle that case using `isFetching`?

private-templates-service/client/src/components/SideBar/SideBar.tsx
```diff
@@ -98,7 +98,7 @@ const navMenuLinks: INavLinkGroup[] = [
       },
       {
         name: "New Template",
-        url: "/designer",
+        url: "/designer/newcard/1.0",
```
guptaharsh25: can we extract this into `const newTemplateURL = "/designer/newcard/1.0"` and reference that here and on line 158 in this file? Just so that we don't break anything by accident if this URL ever changes.

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
nnmkhang: Kody do you know a way to trigger the call on the falling edge of isFetching? that is what we are trying to do

private-templates-service/client/src/components/NavBar/NavBar.tsx
nnmkhang: I will create another ticket to address this thx

nnmkhang:
> Quick idea that can be included in this PR: When the save modal opens can we populate the card name textfield with the current name of the template? I believe it'll always be "Untitled Card".
> 
> Not necessary but right now it looks like this:
> image
> 
> nav bar implies there is a name attached to this card but save modal doesn't reflect that

There is no opportunity to change the name for the first time when creating a new card until the user enters the save modal. That is why the name in the text field is always 'Untitled Card'

private-templates-service/client/src/components/Common/SpinnerModal/styled.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+import { Spinner } from 'office-ui-fabric-react/lib/Spinner';
+import { COLORS } from '../../../globalStyles';
+
+export const CenteredSpinner = styled(Spinner)`
+  margin: auto;
+`;
+
+export const Dot = styled.div`
```
guptaharsh25, line 9: What does this do? I saw that it wraps the spinner.

guptaharsh25: Is is just a white circular background that the spinner can sit on top off?

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -60,18 +66,42 @@ interface State {
...
+
+  componentDidUpdate() {
+  if (this.props.location.pathname == '/designer/newcard/1.0' && this.props.templateID && this.props.version) {
+    this.props.history.replace('/designer/' + this.props.templateID + '/' + this.props.version);
+  }
+    this.toggleSpinner(this.props.isFetching);
+    if (this.props.templateJSON){
+      designer.setCard({...this.props.templateJSON});
+    }
+  }
```
guptaharsh25, line 104: nit: formatting (whitespace) seems off here, but that might just be Github

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -60,18 +66,42 @@ interface State {
...
-    this.state = { isSaveOpen: false };
+    this.state = {isSaveOpen: false };
+    if(this.props.match.params.uuid !== "newcard"){
+      this.props.getTemplate(this.props.match.params.uuid);
+    }
   }
 
   toggleModal = () => {
     this.setState({ isSaveOpen: !this.state.isSaveOpen });
   }
```
guptaharsh25, line 88: is this used anywhere? I didn't see it. If not line 80 `this.state = {isSaveOpen: false };` and the State interface can be removed as well.

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -99,7 +129,7 @@ class Designer extends React.Component<DesignerProps, State> {
     designer.monacoModuleLoaded(monaco);
 
     if (this.props.templateJSON) {
-      designer.setCard(this.props.templateJSON);
+      designer.setCard({...this.props.templateJSON});
```
guptaharsh25, line 132: why did we have to do this again?

private-templates-service/client/src/components/Designer/SaveModal/SaveModal.tsx
```diff
@@ -131,7 +131,7 @@ class SaveModal extends React.Component<Props, State> {
               </Container>
               <InfoWrapper>
                 <StyledH3>{CARDNAME}</StyledH3>
-                <StyledTextField onChange={this.onChange} />
+                <StyledTextField onChange={this.onChange} placeholder = {UNTITLEDCARD} />
```
guptaharsh25, line 134: `placeholder` sets the greyed out unselectable text inside the textfield, `defaultValue` sets the initial value of the textfield.

IMO, `defaultValue={UNTITLEDCARD}` and `placeholder={"My Card"}`, something like that.

private-templates-service/client/src/store/page/types.ts
```diff
@@ -11,7 +11,8 @@ export enum ModalState {
   Version,
   EditName,
   Delete,
-  Save
+  Save,
+  Spinner
```
guptaharsh25, line 15: I don't think you use this anymore.

private-templates-service/client/src/components/Common/SpinnerModal/styled.tsx
nnmkhang: yes

private-templates-service/client/src/components/Designer/Designer.tsx
nnmkhang: This is for when the designer is opened the second time. It has an issue with binding with designer.setCard(). This makes sure that we create a new object from the props and not use any weird binding from setCard()

nnmkhang:
> good work comrade
Alt Text

private-templates-service/client/src/components/Common/SpinnerModal/styled.tsx
```diff
@@ -0,0 +1,16 @@
...
+  height: 30px;
+  width: 30px;
+  background-color: ${COLORS.WHITE};
+  border-radius: 50%;
```
kodyang, line 13: nit
 ask naomi how the spinner background should look, I think its more standard to have a square background.

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const mapStateToProps = (state: RootState) => {
+  return {
+    isFetching: state.currentTemplate.isFetching
+  }
+}
+
+interface Props {
+  isFetching: boolean;
+}
```
kodyang, line 16: Not used. remove

private-templates-service/client/src/components/Common/SpinnerModal/SpinnerModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <BackDrop>
+        <Dot>
+          <CenteredSpinner size={SpinnerSize.large}/>
+        </Dot>
+      </BackDrop>
+    );
+  }
+}
+
+export default connect(mapStateToProps)(SpinnerModal);
```
kodyang, line 34: `export default SpinnerModal`

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -38,39 +40,57 @@ const mapDispatchToProps = (dispatch: any) => {
...
-    this.state = { isSaveOpen: false };
+    if(this.props.match.params.uuid !== "newcard"){
+      this.props.getTemplate(this.props.match.params.uuid);
+    }
+  }
+
+  toggleSpinner = (isFetching: boolean) => {
+    if(isFetching){
+      this.props.openModal(ModalState.Spinner)
+    }
```
kodyang, line 83: openModal no longer triggering spinner modal. delete

private-templates-service/client/src/components/Designer/Designer.tsx
```diff
@@ -38,39 +40,57 @@ const mapDispatchToProps = (dispatch: any) => {
...
   }
 
-  toggleModal = () => {
-    this.setState({ isSaveOpen: !this.state.isSaveOpen });
+  componentDidUpdate() {
+    if (this.props.location.pathname == '/designer/newcard/1.0' && this.props.templateID && this.props.version) {
+      this.props.history.replace('/designer/' + this.props.templateID + '/' + this.props.version);
+    }
+    this.toggleSpinner(this.props.isFetching);
```
kodyang, line 90: remove

private-templates-service/client/src/components/NavBar/NavBar.tsx
```diff
@@ -101,11 +101,11 @@ const NavBar = (props: NavBarProps) => {
       return (
         <Banner>
           <MobileBanner>
-            <StyledLogo aria-label={STRINGS.LOGO_DESCRIPTION} src={Logo} />
-            <Styledh1>{(props.templateID === "" && STRINGS.UNTITLEDCARD) || props.templateName}</Styledh1>
-            {props.templateID !== "" && <EditButton onClick={editName} iconProps={{ iconName: 'Edit' }} />}
+          <StyledLogo aria-label={STRINGS.LOGO_DESCRIPTION} src={Logo} />
```
kodyang, line 104: nit: tab

PR #200, opened 2020-03-31
Add skip link to main content

beervoley:
This pr addresses [BUG34224](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34224) to create a skip to main content link to address accessibility.
The link won't be visible unless it receives a focus and it will get the focus if the user clicks **tab** while focusing on the **Dashboard** link on the SideBar. The link will change the user focus to RecentlyEdited.
image

### DIFF
--- a/private-templates-service/client/src/components/Common/SkipLink/styled.tsx
+++ b/private-templates-service/client/src/components/Common/SkipLink/styled.tsx
@@ -7,16 +7,14 @@ position: absolute;
 left: -999px;
 top: -999px;
 &: focus {  
-    padding:0 25px 0 25px;
+    font-size: 0.8rem;
+    padding:5px 25px 5px 25px;
     left:0;
     top:75px;
     margin-left: 40px;
-    color: black;
+    color: white;
     outline: none;
-    background: white;
+    background: ${COLORS.BLUE3};
     text-decoration: none;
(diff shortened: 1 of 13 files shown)


private-templates-service/client/src/store/skiplink/actions.ts
```diff
@@ -0,0 +1,8 @@
+import { SET_SKIP_LINK_CONTENT_ID, SkipLinkAction, SkipLinkState } from "./types";
```
turalf: Redundant import? SkipLinkState

private-templates-service/client/src/components/Common/SkipLink/styled.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from "styled-components";
+import { COLORS } from "../../../globalStyles";
+
+export const SkipLinkContainer = styled.a`
+display: block;
+position: absolute;
+left: -999px;
+top: -999px;
```
kodyang: Is this how it is supposed to be implemented or did you just do this to make it hidden

beervoley: I just did it to move it off screen

PR #203, opened 2020-04-01
Make version modal scrollable

gracez72:
Ticket: [TICKET-34327](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/11?workitem=34327)

This PR makes the version lists scrollable when there are many template versions. 

It also contains a verticalScroll function for a custom Scroller class, [Seva's all templates branch](https://github.com/microsoft/adaptivecards-templates/blob/users/vsevolod/all-templates-page/private-templates-service/client/src/utils/AllCardsUtil/AllCardsUtil.ts) will also have a similar class with horizontalScroll. These changes will be combined to single Scroller class in a future PR/when the later branch is merged to dev. 

smallscroll

bigscroll

### DIFF
--- a/private-templates-service/client/src/utils/Scroller.tsx
+++ b/private-templates-service/client/src/utils/Scroller.tsx
@@ -1,0 +1,17 @@
+
+export class Scroller {
+  timeStamp: number;
+  constructor() {
+      this.timeStamp = 0;
+  }
+  // TODO: decide which scroll feels more natural
+  verticalScroll = (e: React.WheelEvent) => {
+      e.preventDefault();
+      let item = e.currentTarget;
+      if (e.timeStamp - this.timeStamp < 1000) {
+          return;
+      }
+      this.timeStamp = e.timeStamp;
+      item.scrollTop += 10 * e.deltaY;
+  }
+}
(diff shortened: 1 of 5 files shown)


kodyang:
Can we get rid of the rightmost scrollbar 
image

private-templates-service/client/src/utils/Scroller.tsx
```diff
@@ -0,0 +1,17 @@
...
+  verticalScroll = (e: React.WheelEvent) => {
+      e.preventDefault();
+      let item = e.currentTarget;
+      if (e.timeStamp - this.timeStamp < 1000) {
+          return;
+      }
+      this.timeStamp = e.timeStamp;
+      item.scrollTop += 10 * e.deltaY;
+  }
+}
```
kodyang, line 17: I think its best to avoid our own scroll logic, though that is a decision for the team to make. We can leave this in for now.

gracez72: I agree, I ran into some issues using the fabric ui scrollable and based this modal after Seva's changes for the tags scolling on the all cards page, I'll see if I can find another way to make it work without the scroll logic in a later PR

PR #208, opened 2020-04-04
Users/nnmkhang/sort filter

nnmkhang:
IGNORE, Reformating the way this is done! 

[AB#34450](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34450) 
[AB#34451](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34451)

Overview of this change:
- Adding Backend support for querying cards based on tags and on names. 
- Adding Support in backend and frontend for sorting tags by: alphabetical, date updated and date created

Note:
- The frontend api calls for filter have been updated but the backend has not yet been configured for it.


private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -32,8 +32,10 @@ export class MongoDBProvider implements StorageProvider {
         })
       };
     }
-
-    return templateQuery;
+    if(templateQuery.name && templateQuery.tags){
+      return {$or: [{name:templateQuery.name},{tags:templateQuery.tags}]}
```
kodyang, line 36: nit: please add space after commas
`return {$or: [{name:templateQuery.name}, {tags:templateQuery.tags}]}`

private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -32,8 +32,10 @@ export class MongoDBProvider implements StorageProvider {
         })
       };
     }
-
-    return templateQuery;
+    if(templateQuery.name && templateQuery.tags){
+      return {$or: [{name:templateQuery.name}, {tags:templateQuery.tags}]}
+    }
+    return templateQuery
```
beervoley, line 38: Or condition is correct but partial name/tag search is not going to return correct results. If you look a little bit up in this function, you'll see how partial search is done for name and tags. Try to put that in. I might be mistaken though and your approach will work just fine.
The **$all** operator for tags above is used to find templates that have all the specified tags inside (they might have more than that). https://docs.mongodb.com/manual/reference/operator/query/all/

PR #211, opened 2020-04-06
Add basic deploy to azure button

gracez72:
**Tickets**: [TASK-34368](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34368), [TASK-34449](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34449), [TASK-34414](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/12?workitem=34414)

### Summary of Changes
**client/**
Removed Config file and react environment variables from the client code. The client now requests the environment variables (app id and redirect uri) from the backend when the page is first loaded. 
**server/**
Added config endpoint for client to request environment variables. 

**adaptivecards-templating-service/**
Added config endpoint for client to request environment variables. Previous string token replaced variables are replaced with environment variables. 
Published to [npm registry](https://www.npmjs.com/package/adaptivecards-templating-service)
**adaptive-templating-service-typescript-node/**
Published to [npm registry](https://www.npmjs.com/package/adaptive-templating-service-typescript-node). 

**Docker**
Uploaded docker image to ACMSRegistry on Azure Container Registry with most recent dev changes and changes in this branch. 

### Testing these Changes
1. Go to [private-templates-service's README](https://github.com/microsoft/adaptivecards-templates/tree/users/grzhan/docker/private-templates-service). Click the blue 'Deploy to Azure' button. 
2. Select the subscription, resource group, and server farm you wish to deploy your website under.
3.  **If you already have an existing AAD app registration you want to use, skip this step**. Open Azure portal and open Azure Active Directory and select App registrations. Create a new registration with your desired site url. 
4. With your existing or newly created AAD app registration, double check that your new website url is listed as a redirect URI.
5. Enter in your existing mongodb instance's connection string, and the credentials to access the acmsregistry.azurecr.io/acmsimage docker image and deploy. 

The deployment process will take around 10 minutes on the basic plan, you can monitor the progress under Monitor/Log Stream or under Container settings. If you find that it is taking over 10 minutes, restart the web app. 

### Running locally
1. If you already have node_modules installed under server or client, run `npm uninstall adaptivecards-templating-service` under server and `npm uninstall adaptive-templating-service-typescript-node` under client. 
2. Follow the instructions [here](https://github.com/microsoft/adaptivecards-templates/tree/dev/private-templates-service) to run the server and client locally. 

[...]

PR #212, opened 2020-04-06
Move deployment to Azure Dev Ops

daniellemott:
NOTE: This PR conflicts with https://github.com/microsoft/adaptivecards-templates/pull/211. If it goes in first, I will have to make some changes. 

This PR moves our deployment process to Azure Dev Ops. It also enables monitoring of our dev and stage websites by connecting the websites to Azure App Insights.

Co-author: Grace Zhang

Frontend changes (including the client SDK):
- Frontend environment variables names are prefixed with "REACT_APP".
- The frontend environment variable values are assigned in the corresponding pipeline on Azure Dev Ops.

Backend changes:
- Backend environment variable values are assigned in the Application Settings which is found under the Configuration tab of the corresponding App Service 
- The DB connection string is stored in Azure Key Vault for compliance reasons. The DB connection string environment variable is instantiated under the Application Settings for the App Service but it's value is a reference to the Azure Key Vault entry. 

All environment variables in the frontend and backend are referenced using the proccess.env.VAR_NAME format. 

Our pipelines are found under the ACMS folder: https://dev.azure.com/microsoft/AdaptiveCards/_build?view=folders


daniellemott:
> how does this change our dev processes ? do we just merge as usual and it will deploy to dev

If you merge to dev or stage, it will just deploy. The biggest change would be that when a PR is created, you have to go the Azure Pipelines to see the details of the build and tests. You will only be able to see if they pass or fail via GitHub.

PR #213, opened 2020-04-07
Change owner from Template field to Template Instance field

kodyang:
## Description
Cards have many versions. It makes sense each individual version has an `Author`, instead of just the card itself. 

[AB#34129](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34129)


private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -410,6 +405,7 @@ export class TemplateServiceClient {
           templateInstance.state = templateInstance.state !== TemplateState.deprecated && templateInstance.state !== TemplateState.draft &&
             instance.state !== TemplateState.draft ? TemplateState.draft :
             templateInstance.state || instance.state;
+          templateInstance.author = instance.author; // author should not be able to be overwritten
```
daniellemott, line 408: If a template is published and someone who is not the author of that version of the template unpublished it, then will the original author be maintained?

kodyang: yes. that is the whole idea.

I think if a version goes through changes, the original author should remain the same through all changes in that specific version.

kodyang:
Hey Danielle, great question. 

I was thinking that as well, but we actually can't use `author` where we use `lastEditedUser`, since it is our db user id rather than the azure oID.

We can discuss with grace and pm + design if we want to change this.

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
daniellemott: My one concern is that other people can change the template version. If they do change the version, it's as if they're making changes under the author's name, instead of their own name. 

This may be simply a limitation of constructing the owner system this way. I think in an ideal world, we would be keeping track of all people that have changed the template. Something like what GitHub does: 
image
But, I'm sure that's out of scope of our internship :(

daniellemott:
> Hey Danielle, great question.
> 
> I was thinking that as well, but we actually can't use `author` where we use `lastEditedUser`, since it is our db user id rather than the azure oID.
> 
> We can discuss with grace and pm + design if we want to change this.

Sounds good!

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -431,14 +427,20 @@ export class TemplateServiceClient {
     } else {
       templateInstances.push(templateInstance);
     }
-    let userResponse = await this._getUserID(authId);
-    if (!userResponse.success) return { success: false, errorMessage: userResponse.errorMessage };
+
+    // Re-create entire author array
```
gracez72, line 431: Do we need to recreate? Can use authorsList from existingTemplate and add the userResponse.result to the array if it doesn't already exist

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
```diff
@@ -933,6 +940,16 @@ export class TemplateServiceClient {
     }
     template.instances = templateInstances;
     template.deletedVersions?.push(...versionList);
+
+    // Re-create entire author array
+    let authorsList: string[] = [];
+    for (let instance of templateInstances) {
```
gracez72, line 946: Can move this check to the above template.instances loop

private-templates-service/adaptivecards-templating-service/src/TemplateServiceClient.ts
kodyang: i agree, it can be a little misleading. however, the `author` is not actually displayed in `templateinfo`, rather we show the `lastediteduser`. author is not really a visible field so i think its not a concern.

kodyang: The designs are adjusted to show `Last Edited By` rather than `author`

PR #219, opened 2020-04-09
Add tag support to All Cards page

beervoley:
This PR addressed:
[TASK-34380](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34380)
[TASK-34460](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34460)
[TASK-34461](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34461)
[TASK-34501](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34501)
[TASK-34138](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34138)

What not to review:
`api.ts` file - it was auto generated based on `swagger.yaml`

What has been done:
1) Displaying all tags as a horizontal list on top of the All Cards Page
2) Tags are clickable: when tag(s) is(are) clicked -> its(their) state changes to selected and the page displays only the templates that have chose tag(s).
3) Selected and not selected tags have different styles.
4) If opening a Template Preview from All Cards Page and then clicking `Back` or `All Cards` in the SideBar -> all the chosen tags will remain the same and won't be reset. To reset the tags -> click any other link on the portal and come back to All Cards.

What has not been done:
1) Star Icon inside the tag to mark tag as favourite was not implemented.
2) If there are more tags than fit on the screen there should be `+ number of tags hidden` button to expand all the tags on click. Right now tags are (horizontally) scrollable if they don't all  fit on the screen at once.
3) If after opening Template Preview from All Cards page one(many) of the chosen tags was(were) deleted -> no reload happens on the page. It will try to display the templates that match the chosen tags and if there are none page will render `Nothing to display`. I will mark it as a bug and we can fix it later (it's moderate amount of work and not high priority).

Screenshot of the page states:

No tags selected:
image

Some tags selected:
image

Interaction with the page GIF:
out16


private-templates-service/adaptivecards-templating-service/src/storageproviders/MongoDBProvider.ts
```diff
@@ -28,7 +28,7 @@ export class MongoDBProvider implements StorageProvider {
     if (query.tags && query.tags.length) {
       templateQuery.tags = {
         $all: clone(query.tags).map(x => {
-          return x.toLocaleLowerCase();
+          return new RegExp(`^${x}$`, "i");
```
beervoley, line 31: This change will be added with https://github.com/microsoft/adaptivecards-templates/pull/208. I've added it here for correct functioning of All Cards in case this PR gets merged first.

private-templates-service/client/src/components/AllCards/AllCards.tsx
```diff
@@ -22,10 +26,16 @@ import requireAuthentication from "../../utils/requireAuthentication";
 // Strings
 import { ALL_CARDS_LIST_VIEW, ALL_CARDS_GRID_VIEW, ALL_CARDS, ALL_CARDS_TITLE } from "../../assets/strings";
 import TagList from "./TagList";
+import { COLORS } from "../../globalStyles";
+
+
```
turalf, line 31: Intended extra empty lines? Do you have a code policy checker on the pipeline for this kind of stuff?

private-templates-service/client/src/components/AllCards/AllCards.tsx
```diff
@@ -38,23 +48,79 @@ const mapDispatchToProps = (dispatch: any) => {
...
 };
 interface Props extends RouteComponentProps {
-  templates: AllTemplateState;
   setPage: (currentPageTitle: string, currentPage: string) => void;
   setSearchBarVisible: (isSearchBarVisible: boolean) => void;
   toggleView: (viewType: ViewType) => void;
+  getAllTags: () => void;
+  addSelectedTag: (tag: string) => void;
+  removeSelectedTag: (tag: string) => void;
+  clearSelectedTags: () => void;
```
turalf, line 73: are those intended for removing a single tag vs removing all of them? would it be better to have only one and if parameter is not given remove all of them?

private-templates-service/client/src/components/AllCards/AllCards.tsx
```diff
@@ -22,10 +26,16 @@ import requireAuthentication from "../../utils/requireAuthentication";
...
+import { COLORS } from "../../globalStyles";
+
+
 
 const mapStateToProps = (state: RootState) => {
   return {
-    isSearch: state.search.isSearch
+    isSearch: state.search.isSearch,
+    tags: state.allTags,
+    selectedTags: state.selectedTags,
```
turalf, line 37: maybe a single "tags" property and have a boolean flag "isSelected"?

private-templates-service/client/src/components/AllCards/AllCards.tsx
```diff
@@ -38,23 +48,79 @@ const mapDispatchToProps = (dispatch: any) => {
...
+class AllCards extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.props.setPage(ALL_CARDS_TITLE, ALL_CARDS);
     this.props.setSearchBarVisible(true);
+    this.setSelectedTags();
+  }
+  
+  setSelectedTags = (): void => {
+    if(this.props.page.currentPage === "Template" && this.props.page.previousPage === ALL_CARDS) {
```
turalf, line 92: is it intended that tags are selected only if Template page is reached from ALL_CARDS?

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -123,7 +123,7 @@ class Dashboard extends React.Component<Props> {
     }
   }
   selectTemplate = (templateID: string) => {
-    this.props.history.push("preview/" + templateID);
+    this.props.history.push("/preview/" + templateID);
```
guptaharsh25, line 126: curious, is this a necessary change?

private-templates-service/client/src/store/page/reducer.ts
```diff
@@ -13,12 +13,14 @@ const initalState: PageState = {
 }
 
 export function pageReducer(state = initalState, action: PageAction): PageState {
+  // console.log(state);
```
guptaharsh25, line 16: remove

private-templates-service/client/src/store/page/reducer.ts
```diff
@@ -13,12 +13,14 @@ const initalState: PageState = {
...
   switch (action.type) {
     case NAVIGATION:
       return {
         ...state,
         currentPageTitle: action.currentPageTitle,
-        currentPage: action.currentPage
+        currentPage: action.currentPage,
+        previousPage: state.currentPage
```
guptaharsh25, line 23: why do you need previous page?

private-templates-service/client/src/store/templates/actions.ts
```diff
@@ -23,7 +23,24 @@ export function failGetAll(error: IncomingMessage): AllTemplateAction {
     error: error
   }
 }
-
+export function getTemplatesByTags(tags?: string[]): (dispatch: any, getState: () => RootState) => void {
```
guptaharsh25, line 26: do we need this as a separate action? can we modify the existing `getAllTemplate` action to accept an optional param called tags?

private-templates-service/client/src/store/page/reducer.ts
guptaharsh25: there should be a way to access previous page with react router

private-templates-service/client/src/components/AllCards/AllCards.tsx
```diff
@@ -38,23 +48,79 @@ const mapDispatchToProps = (dispatch: any) => {
...
+    this.setState((state) => {
+      if(state.selectedTags.includes(tag)) {
+        this.props.removeSelectedTag(tag);
+        return {selectedTags: state.selectedTags.filter((selectedTag: string) => selectedTag !== tag)}
+      } else {
+        this.props.addSelectedTag(tag);
+        return {selectedTags: state.selectedTags.concat(tag)}
+      }
+    });
+  }
```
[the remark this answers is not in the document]
beervoley: it doesn't return an object: the function that returns an object is passed in setState as argument

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
beervoley: yep, it's a necessary change. If there's no slash in front of `preview` than when router does the routing it will use relative address instead of absolute. Example: /template/all => after push becomes /template/preview/..... instead of /preview/...

private-templates-service/client/src/store/page/reducer.ts
beervoley: there's no way to access previous page either with history or router

private-templates-service/client/src/store/templates/actions.ts
beervoley: yes, we can do that

private-templates-service/client/src/components/AllCards/AllCards.tsx
beervoley: It's other way around. When we look at `currentPage` in constructor it actually will be `previousPage` since update from `setPage(ALL_CARDS..)` is not in the current props. So, when I check for `currentPage === "Template"` I'm actually asking if previous page was `Template` so I can restore state of selected templates.

private-templates-service/client/src/components/AllCards/AllCards.tsx
beervoley: I can't come up with naming of the method that wouldn't make the purpose of the function confusing. If you have any idea, please let me know!

private-templates-service/client/src/components/AllCards/AllCards.tsx
beervoley: eslint should take care of it:)
removed

private-templates-service/client/src/components/AllCards/TagList/TagList.tsx
```diff
@@ -26,9 +29,13 @@ class TagList extends Component<TagListProps> {
   }
   render() {
     return (
+      // The div tag is required to prevent tags container from collapsing on window resize
+      // due to TagsContainer overflows being set to hidden.
+      <div>
```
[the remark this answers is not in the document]
beervoley: Done :)

private-templates-service/client/src/components/AllCards/AllCards.tsx
turalf: The naming looks fine. I meant, if they are doing the same thing: i.e delete a selected tag or delete all tags, there could be a single method do both depending on the arguments.

PR #221, opened 2020-04-09
Profile Picture Bug + Long Title Bug

kodyang:
## Description

Two bug fixes:
Profile picture does not show on first load. [AB#34331](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34331)
- Fixed by requiring graph access token in requireAuthentication HOC
Long titles do not show properly. [AB#34532](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34532)
-  Fixed by adding ellipses with css


private-templates-service/client/src/components/Dashboard/Dashboard.tsx
```diff
@@ -182,8 +182,8 @@ class Dashboard extends React.Component<Props> {
                   <TemplateList
                     onClick={this.selectTemplate}
                     templates={recentlyViewedTemplates}
-                    displayComponents={{author: true, status: true, dateModified: true, templateName: true, version: false}}
-                  ></TemplateList>
```
turalf, line 186: was this change necessary? or it is just fixing the styling?

private-templates-service/client/src/utils/requireAuthentication/requireAuthentication.tsx
```diff
@@ -5,23 +5,27 @@ import { UserType } from '../../store/auth/types';
...
+    graphAccessToken: state.auth.graphAccessToken,
     user: state.auth.user
   };
 };
 
 const requireAuthentication = <P extends object>(Component: React.ComponentType<P>) => {
   return class AuthenticatedComponent extends React.Component<P & AuthProps> {
 
-    isAuth = () => this.props.isAuthenticated;
+    isAuth = () => this.props.isAuthenticated && this.props.graphAccessToken !== undefined;
```
daniellemott, line 28: So the problem was that the graph access token wasn't in the redux store when the action to get names + profile pictures were dispatched?

kodyang: yup

private-templates-service/client/src/components/Dashboard/Dashboard.tsx
kodyang: just styling

PR #231, opened 2020-04-13
"Tall" adaptive card bugs

daniellemott:
[AB#34519](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/13?workitem=34519) "Tall" adaptive card bugs

This PR fixes the following issues:
- When the Publish modal is opened with a tall adaptive card, the styling is "broken"
image
- When the Template Info modal is opened with a tall adaptive card, the adaptive card overflows out of the border. This is noticeable in the grey line that divides the left and right side of the Template Info modal 
image


private-templates-service/client/src/components/Common/PublishModal/styled.tsx
```diff
@@ -56,6 +56,11 @@ export const AdaptiveCardPanel = styled.div`
   margin-bottom: 16px;
 `;
 
+export const CardWrapper = styled.div`
+  height: 220px;
+  overflow: hidden;
+`;
+
```
guptaharsh25, line 63: Can we see what this looks like?

daniellemott:
> Can you provide screenshots of the fix?
> 
> I think the best experience for the user is if tall cards are scrollable.

I asked Naomi about scrolling for tall cards in the various modals (publish, unpublish, etc.) but she preferred that we just cut them off.

PR #238, opened 2020-04-14
Opt-in telemetry + fix frontend telemetry bug

daniellemott:
[AB#34089](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/Adaptive%20Cards/Intern%20GitHub/Y20-VW/13?workitem=34089)

This PR enables users to pass in their instrumentation key for their own instance of Azure App Insights when using the `Deploy to Azure` button. 

image

This PR also fixes a bug that prevents our frontend telemetry from working. The backend was using the variable name `appInsightInstrumentationKey`, but it should have been labelled `appInsightsInstrumentationKey`.


private-templates-service/azuredeploy.json
```diff
@@ -2,145 +2,149 @@
...
+      "type": "String"
+    },
+    "telemetry_opt_in": {
+      "type": "string",
+      "allowedValues": [
+        "Yes",
+        "No"
+      ]
+    },
+    "application_insights_instrumentation_key": {
```
gracez72, line 40: Can you update the documentation to include where to find this value in Azure portal?

kodyang:
lgtm, though i will approve once grace approves. not my domain

PR #265, opened 2020-04-21
Fix scroll on preview

beervoley:
This PR addresses [34654](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/34654) to fix main window in template preview becoming scrollable due to too many versions displayed in the Recent Releases.
I've also fixed the scrolling with touchpad issue and made shadow of NavBar visible.

Before:
image

After:
image


beervoley:
> In terms of vertical scroll, the TemplateInfo page looks great!
> 
> In terms of horizontal scroll, I think the Tags table needs to be a fixed width. If too many tags are added, the right pane of the window scrolls horizontally:
> image

Fixed it :)

PR #273, opened 2020-04-22
Set overflow-behaviour for Designer Page

guptaharsh25:
[AB#34528](https://microsoftgarage.visualstudio.com/002806e2-ebaa-4672-9d2e-5fe5d29154ef/_workitems/edit/34528)

This will help with ensuring there's no scrolling on the designer page.

### DIFF
--- a/private-templates-service/client/src/components/Designer/styled.tsx
+++ b/private-templates-service/client/src/components/Designer/styled.tsx
@@ -4,7 +4,7 @@ export const OuterDesignerWrapper = styled.main`
   display: flex;
   flex-direction: column;
   height: 100%;
-  overflow-behaviour: contain;
+  overscroll-behaviour: contain;
 `;
 
 export const DesignerWrapper = styled.div`
--- a/private-templates-service/client/src/components/Designer/styled.tsx
+++ b/private-templates-service/client/src/components/Designer/styled.tsx
@@ -4,6 +4,7 @@ export const OuterDesignerWrapper = styled.main`
   display: flex;
   flex-direction: column;
   height: 100%;
+  overflow-behaviour: contain;
 `;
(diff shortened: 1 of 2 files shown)


private-templates-service/client/src/components/Designer/styled.tsx
```diff
@@ -4,6 +4,7 @@ export const OuterDesignerWrapper = styled.main`
   display: flex;
   flex-direction: column;
   height: 100%;
+  overflow-behaviour: contain;
```
beervoley, line 7: I think it's a typo: `overscroll-behavior` - it should be

guptaharsh25: my bad, fixed

private-templates-service/client/src/components/Designer/styled.tsx
```diff
@@ -4,7 +4,7 @@ export const OuterDesignerWrapper = styled.main`
   display: flex;
   flex-direction: column;
   height: 100%;
-  overflow-behaviour: contain;
+  overscroll-behaviour: contain;
```
gracez72, line 7: is it supposed to be behavior?

guptaharsh25: i think so, `overscroll-behavior: contain;` sets the inner elements to be unscrollable.
https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior

gracez72: I mean without the u in behaviour!

guptaharsh25: haha whoops, fixed

